PR #61, opened 2026-07-12
ALT-01: datum-qualified altitude — the tape identifies its reference, nothing substitutes

luofang34:
Implements #16. **SIM / NOT FOR FLIGHT** — the simulator's local-relative altitude remains the only real source; this PR makes that fact visible at the type level, on the wire, and on the tape, and makes every conventional reference fail closed until a real source exists.

## Design

- **`AltitudeClass` / `AltitudeReference`** (new `altitude` module in `pilotage-instrument-state`): local-relative (with `OriginId`), barometric indicated (with the applied setting), pressure, geometric MSL (with `GeoidModelId`), AGL. Unknown wire bytes decode to `Unknown` and fail — a reference is never guessed (`AltitudeClass::from_u8`, fail-closed round-trip tested).
- **`PanelData.alt_ft` is gone.** Panels consume `ResolvedAltitude { value_ft, class, origin, setting_mismatch, bug_compatible }` — there is no unqualified altitude left to consume (acceptance: "every panel altitude identifies its reference at the type level").
- **No fallback, ever**: a barometric class without its sample *and* source-applied setting resolves `Failed` with a typed `SourceAbsent` integrity reason and a quiet-zero value — the NED-derived number cannot leak into a barometric tape (`baro_class_without_a_source_fails_and_never_uses_ned`). Geometric MSL additionally requires a declared geoid model. Local NED z feeds *only* the local-relative class.
- **Visible REL**: the tape paints its reference label under the value box — REL in amber, BARO/STD/MSL/AGL in white, red REF beside a failed tape for an unknown reference (`local_relative_tape_is_labelled_rel`, pinned into the PFD frame hash).
- **Setting applied vs shown**: the setting readout states its own semantics — cyan value when applied (barometric), `STD` (pressure), grey `SET <value>` when displayed but *not* applied (every other reference). Pilot-selected setting (`selections.baro_sel_hpa`) is separate from the sensed applied setting (`AirData`); disagreement beyond 0.5 hPa raises the amber `BARO SEL` flag and is never averaged (`setting_mismatch_is_flagged_and_agreement_is_not`).
- **Selection compatibility is full datum identity — class, origin, model, and setting agreement**: the bug and selection readout render only when `selections.altitude_sel_class` matches the displayed class; an incompatible selection renders the amber `SEL REF` marker instead of a plausible number (`bug_requires_matching_class_before_identity_is_even_considered`, `incompatible_selection_never_renders_a_plausible_number`).
[...]

Issue #21, opened 2026-07-11
NAV-01: separate heading from attitude yaw and type magnetic/true references


luofang34:
### Problem

The current resolver treats attitude Euler yaw as heading. It is local-NED orientation, is ill-conditioned near vertical pitch, and has no magnetic/true reference. HSI heading, track, course, selected bug, bearings, and wind can therefore be mixed without a validated reference conversion.

### Scope

- Add an independent heading sample; operational heading must not be implicit quaternion yaw.
- Type all horizontal angular quantities as magnetic or true: heading, track, selected heading, course, bearing, and wind direction.
- Convert only with a valid magnetic-variation sample whose sign convention, source, epoch, and freshness are explicit.
- Centralize and test the variation convention and shortest-angle arithmetic.
- Resolve every HSI quantity into one compatible reference or visibly fail/degrade it.
- Simulator local-NED yaw may be admitted only as explicitly true/local simulation heading.
- Near vertical attitude, independent heading remains independent; an undefined source produces a flag rather than fabricated yaw.
- Update schema/ABI and display reference annunciation.

### Non-goals

- AHRS/magnetometer fusion.
- World Magnetic Model lifecycle.
- Grid/polar navigation.
- Guidance algorithms or certification claims.

### Acceptance criteria

- [ ] PanelData exposes no unreferenced heading.
- [ ] HSI heading, track, bug, course, bearing, and wind use a common resolved reference or visibly fail.
- [ ] Quaternion yaw is not the implicit operational heading.
- [ ] True/magnetic conversion is centralized.
- [ ] 359/0 degree wrap is correct.
- [ ] Missing heading cannot leave a plausible frozen compass rose.

### Deterministic verification

Test 359 versus 1 degree, both signs of variation across north, agreement/conversion/missing/stale/mismatch, independent heading at pitch 89/90/91 and inverted attitudes, missing heading with valid attitude, and unknown-reference fail-closed ABI behavior.

### Dependencies

AV-01 and VAL-01. Supplies typed inputs for SRC-01. Coordinate PanelData changes with ATT-01.

### Design inputs

[FAA AC 20-181](https://www.faa.gov/documentLibrary/media/Advisory_Circular/AC_20-181.pdf) and [FAA AC 25-11B](https://www.faa.gov/documentlibrary/media/advisory_circular/ac_25-11b.pdf). Completion is not an approval claim.

luofang34:
[...]

PR #63, opened 2026-07-12
NAV-01: typed magnetic/true heading — the rose turns on an independent sample

luofang34:
Implements #21. **Stacked on #61 (ALT-01)** — both revise the state ABI; merge #61 first. **SIM / NOT FOR FLIGHT.**

## Design

- **Quaternion yaw is out of the heading path entirely.** `attitude_geometry` no longer produces yaw; `PanelData.heading_rad` is gone. Heading is `ResolvedHeading { value_rad, reference }` from an independent `HeadingSample` group with its own age and declared-valid bit (acceptance: "PanelData exposes no unreferenced heading", "Quaternion yaw is not the implicit operational heading").
- **References are typed** (`HeadingReference`): Magnetic, True, and `SimLocalTrue` — the sanctioned, *explicitly declared* admission of simulator local-NED yaw, labelled amber `SIM` on the rose so it can never pass for a navigation-grade source. Unknown wire bytes fail closed.
- **One conversion path** (`convert_heading` in the new `heading` module): magnetic/true crossings require a usable `MagneticVariation` sample — east-positive sign convention fixed in the type, source identity mandatory (`VariationSourceId::UNDECLARED` refuses), freshness enforced at the call site. Stale, undeclared, or non-finite variation never rotates the compass. Wrap arithmetic (`wrap_2pi`, `shortest_angle_rad`) is centralized beside it.
- **One reference per rose**: NED-derived true quantities (track, wind) present in the heading sample's reference. On a magnetic rose without usable variation they *fail visibly* rather than mixing references (`magnetic_display_without_variation_degrades_true_quantities`); with variation they convert exactly (`358°` for a true-north track under 2°E).
- **No fabricated rose**: a missing sample resolves `Missing` → HDG red-X, and the cardinal letters do not paint (`missing_heading_leaves_no_plausible_rose`). At pitch 89°/90°/91° and inverted, the independent sample passes through bit-unchanged (`heading_is_independent_of_attitude_through_the_vertical`).
- **State ABI v4 (168 bytes)**: heading value/age/reference and variation value/age/source appended with a second valid-flags byte; the JS writer's fail-safe default writes *no* sample (reference 255, NaN value) — a default heading of zero would be exactly the plausible frozen rose the issue bans.

| Criterion (#21) | Evidence |
|---|---|
| PanelData exposes no unreferenced heading | `ResolvedHeading` is the only heading; field `heading_rad` removed repo-wide |
| HSI quantities share one resolved reference or visibly fail | `presented_true`/`presented_wind` + degrade tests; SIM/MAG/TRU label on the rose |
| Quaternion yaw not the implicit operational heading | yaw dropped from `attitude_geometry`; `missing_heading_resolves_missing_even_with_a_valid_attitude` |
| True/magnetic conversion centralized | `convert_heading` is the single path; callers cannot add/subtract variation themselves |
| 359/0 wrap correct | `wrap_covers_the_359_to_1_degree_boundary_both_ways`, conversions across north both signs |
| Missing heading cannot leave a plausible frozen rose | `missing_heading_leaves_no_plausible_rose` (no cardinals, HDG flag) |
| Independent heading at 89/90/91 and inverted | `heading_is_independent_of_attitude_through_the_vertical` |
| Unknown-reference fail-closed ABI | `unknown_heading_reference_byte_decodes_fail_closed`, `unknown_reference_fails_the_heading` |

## Feeder follow-up (deliberate)

The live browser feeder (`main.js`) does not yet declare the explicit SIM heading — that file is owned by the in-flight HUD-01 lane this wave, and editing it now would tangle the lanes. Until the one-line declaration lands as a follow-up commit, the live HSI honestly flags HDG (fail-safe, not a regression of information: the previous rose was an unlabelled implicit-yaw fabrication). The wasm/state/HSI path is fully proven by the fixture and runtime tests.

## Verification

Lane-scoped gates green: fmt, clippy `-D warnings`, tests (9 new resolution tests, 6 heading-math tests, 3 HSI rendering tests, ABI round-trip/fail-closed), doc, no_std, structure gate (resolve split into `heading_signal` submodule; decode/encode baselines ratcheted down again 132→128 / 105→97). HSI frame hash re-pinned with rationale (rose now driven by the explicit SIM sample); PFD hash byte-identical. JS parity + all touched node suites green; wasm rebuilt.

Refs #21.

### DIFF
--- a/clients/web-instruments/src/tests.rs
+++ b/clients/web-instruments/src/tests.rs
@@ -48,6 +48,7 @@ pub(crate) fn attitude_state() -> AircraftState {
             rates: true,
             position: true,
             velocity: true,
+            ..Default::default()
         },
         ..AircraftState::default()
     }
--- a/clients/web/instruments.js
+++ b/clients/web/instruments.js
@@ -35,8 +35,8 @@ export const EXPECTED_GLYPH_SHA256 =
 const GLYPH_HEADER_LEN = 8;
 const GLYPH_RECORD_LEN = 12;
 const GLYPH_ROWS = 7;
-export const STATE_ABI_VERSION = 3;
-export const STATE_ABI_SIZE_BY_VERSION = Object.freeze({ 1: 120, 2: 128, 3: 144 });
+export const STATE_ABI_VERSION = 4;
+export const STATE_ABI_SIZE_BY_VERSION = Object.freeze({ 1: 120, 2: 128, 3: 144, 4: 168 });
 export const STATE_ABI_SIZE = STATE_ABI_SIZE_BY_VERSION[STATE_ABI_VERSION];
 const MAX_WASM_RENDER_STATUS = 10;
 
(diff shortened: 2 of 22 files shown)


luofang34:
Superseded by the rework PR: the review required typed bug/course references and the live SIM feeder declaration, and the revised #61 ABI made this branch's v4 layout unrebaseable. The successor carries every original test plus the review matrix.

Issue #23, opened 2026-07-11
DYN-01: replace body-yaw-rate proxy with typed turn and slip/skid data


luofang34:
### Problem

The resolver maps body-axis yaw rate r directly to a standard-rate turn indication. At nonzero roll/pitch, body r, heading rate, track rate, and rotation about local vertical are not interchangeable. There is no slip/skid input, and invalid turn data can simply disappear.

### Scope

- Replace the body-rate proxy with a typed turn indication whose basis is explicit, such as heading rate or track rate.
- Prefer an upstream AHRS/navigation estimate with source metadata. Any derivation uses coherent timestamps and circular differencing with documented limits.
- Never silently substitute body r.
- Add typed slip/skid input from an appropriate upstream value such as lateral specific force or normalized coordination output.
- Define body/display axes and signs; missing slip remains missing, never synthesized centered.
- Airframe profile controls whether missing/failed turn/slip requires a visible failure cue.
- Clamp only the pointer; retain the validated over-range value for monitoring.
- Keep turn rate, slip/skid, and aerodynamic sideslip distinct.

### Non-goals

- Sensor fusion, sideslip estimation, autopilot coordination, or mechanical-instrument dynamics.
- Certification claims.

### Acceptance criteria

- [ ] No path labels body r as standard-rate turn.
- [ ] The displayed turn value identifies basis and source status.
- [ ] Missing slip is not shown centered.
- [ ] Sign convention is documented and symmetric.
- [ ] Required failures are explicit.
- [ ] Display saturation preserves the original value for monitoring.

### Deterministic verification

Test body r at rolled/pitched attitude, explicit plus/minus 3 degree/s and over-range turn, 359/1 circular derivation if implemented, symmetric lateral-specific-force mapping, and every status/recovery transition. No sleeps.

### Dependencies

AV-01; derived heading rate also depends on NAV-01. The typed source interface and widget can begin in parallel.

### Design inputs

[FAA AC 20-181](https://www.faa.gov/documentLibrary/media/Advisory_Circular/AC_20-181.pdf) and [FAA AC 25-11B](https://www.faa.gov/documentlibrary/media/advisory_circular/ac_25-11b.pdf).

luofang34:
[...]

PR #69, opened 2026-07-13
DYN-01: typed turn basis and slip/skid — body yaw rate has no path to the display

luofang34:
Implements #23. **SIM / NOT FOR FLIGHT.**

## Design

- **`TurnBasis` makes the proxy unrepresentable**: the vocabulary is heading rate, track rate, or fail-closed Unknown — there is no variant through which a feeder could label body-axis yaw rate as a turn indication, and `attitude_geometry` no longer returns rates at all. An absent dynamics group resolves `Missing`; body `r` cannot leak in at any roll/pitch (`body_yaw_rate_never_reaches_the_turn_indication` pins a 0.3 rad/s body rate at a rolled/pitched attitude producing a Missing turn).
- **Basis and status are displayed**: the cue paints the basis label (HDG/TRK) beside the bar; unknown basis fails the indication.
- **Slip/skid is a typed input**: lateral specific force along body +Y (right positive), ball displaces **opposite** the force — the sign convention is documented on the type and pinned symmetric by both a resolution test and a rendering test. **Missing slip draws brackets and no ball** — never synthesized centered.
- **Clamp only the pointer**: the bar saturates at the ±73 px scale edge while the resolved value retains the over-range rate exactly for monitoring (`pointer_saturates_but_the_resolved_value_does_not`, 20°/s kept bit-exact behind a ±3°/s scale).
- **Profile-controlled failure cues**: `AirframeDisplayProfile.require_dynamics_cue` — when set (simulator profile sets it), a missing/failed turn or slip flags amber TRN / SLIP instead of quietly disappearing; when unset, absence is quiet (both sides tested).
- **Feeder-side derivation with documented limits**: the browser feeder derives heading rate by circular differencing of consecutive explicitly-declared SIM headings on its own coherent clock (dt bounds 5–500 ms; out-of-bounds pairs yield no sample, never a wild or frozen rate; 359°→1° differencing wraps correctly). Derivation lives upstream of the display, never inside it. Slip has no sim source and is never fabricated — the live PFD shows the SLIP flag honestly.
- **State ABI v5 (192 bytes)**: dynamics group (basis byte, turn rate, lateral force, age, valid bits 2/3) with fail-safe defaults, JS writer byte-parity tested including the undeclared-defaults case.

| Criterion (#23) | Evidence |
|---|---|
| No path labels body r as standard-rate turn | `TurnBasis` has no body-rate variant; `body_yaw_rate_never_reaches_the_turn_indication`; resolver never reads `rates_rps` for display |
| Displayed turn identifies basis and source status | basis label rendered (`cue_labels_its_basis`); Sig status drives visibility |
| Missing slip not shown centered | `missing_slip_never_draws_a_centered_ball` (zero circles painted) |
| Sign convention documented and symmetric | module docs + `slip_ball_deflects_opposite_the_lateral_force_symmetrically`, `missing_slip_stays_missing_and_present_slip_is_symmetric` |
| Required failures explicit | `missing_turn_flags_trn_when_the_profile_requires_a_cue` (both profile settings) |
| Saturation preserves original value | `over_range_turn_is_retained_unclamped_for_monitoring` + pointer test |
| ±3°/s exact, symmetric | `plus_minus_three_degrees_reach_the_reference_ticks_symmetrically` (bar geometry at the reference ticks) |
| 359/1 circular derivation | feeder `derivedSimTurn` wraps via (−π, π] differencing; heading-math wrap tests already pin the boundary |
| Status/recovery transitions | staleness, declared-invalidity (turn/slip independent), coherence-skew, non-finite each tested; no sleeps anywhere |

## Verification

Lane gates green: fmt, clippy `-D warnings`, tests across the four instrument crates (10 new resolution tests, 6 rendering tests, ABI round-trip + fail-closed basis, 3 vocabulary tests), doc, no_std, structure gate (resolve/abi split into `dynamics_signal`/`groups` submodules; decode_state baseline ratcheted down 129→125). All node suites pass; wasm rebuilt. **Frame hashes byte-identical** — the demo fixture sits in the unusual-attitude declutter tier where the cue does not paint, so the pinned hashes prove no collateral rendering change; the new drawing paths are covered by the six rendering tests instead.

Refs #23.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,15 @@ jobs:
       - name: renderer backend conformance corpus
         run: node clients/web/scene-conformance.test.mjs
 
+      - name: browser video capture-identity tests (HUD-01)
+        run: node clients/web/video-identity.test.mjs
+
+      - name: browser snapshot-association tests (HUD-01)
+        run: node clients/web/snapshot-association.test.mjs
+
+      - name: browser turn-derivation tests (DYN-01)
+        run: node clients/web/turn-derivation.test.mjs
+
       - name: Setup buf
         if: hashFiles('schemas/**/*.proto') != ''
         uses: bufbuild/buf-action@v1
--- a/clients/web-instruments/src/exports.rs
+++ b/clients/web-instruments/src/exports.rs
@@ -223,7 +223,7 @@ fn derive_alert_events(data: &pilotage_instrument_state::PanelData) -> [AlertEve
             AlertCondition::Heading(NavFault::Unavailable),
         ),
         cond(
-            data.turn_rate_rps.status == SignalStatus::Failed,
+            data.turn.rate_rps.status == SignalStatus::Failed,
             AlertCondition::TurnSlip(DynFault::TurnRateInvalid),
         ),
     ]
--- a/clients/web/instruments.js
+++ b/clients/web/instruments.js
@@ -35,8 +35,8 @@ export const EXPECTED_GLYPH_SHA256 =
 const GLYPH_HEADER_LEN = 8;
 const GLYPH_RECORD_LEN = 12;
 const GLYPH_ROWS = 7;
-export const STATE_ABI_VERSION = 4;
-export const STATE_ABI_SIZE_BY_VERSION = Object.freeze({ 1: 120, 2: 128, 3: 152, 4: 176 });
+export const STATE_ABI_VERSION = 5;
+export const STATE_ABI_SIZE_BY_VERSION = Object.freeze({ 1: 120, 2: 128, 3: 152, 4: 176, 5: 192 });
 export const STATE_ABI_SIZE = STATE_ABI_SIZE_BY_VERSION[STATE_ABI_VERSION];
 const MAX_WASM_RENDER_STATUS = 10;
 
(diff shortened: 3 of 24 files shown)


luofang34:
Review disposition addressed in 09bf319 — the derivation is now measurement-coherent:

- **Derives only on a new accepted measurement identity**: `TurnDerivation` (new `clients/web/turn-derivation.js`) advances only when the heading measurement's stamp presents a serially newer sequence within the same stream (sourceId + sourceIncarnation + sourceEpoch + clock); it is invoked from the render loop but the render loop cannot advance it.
- **dt is the acquisition clock**: differences are over `acquiredAtNanos`, never `performance.now()`; the documented 5–500 ms bound now applies to the measurement interval.
- **Repeated renders re-declare the cached rate** with the measurement's current age — the inflation/oscillation failure mode is structurally gone (`repeated renders re-declare the cached rate unchanged`: sixty frames of one sample, rate bit-identical).
- **Duplicates re-declare; serially older samples are ignored outright**; the state still differences from the newest accepted sample afterward.
- **Resets on session retirement** (`retireSessionPresentation` → `turnDerivation.reset()`) **and every source/incarnation/epoch/clock discontinuity** — a post-reset first sample only seeds, even with an identical stream key and plausible dt (rapid-reconnect test).
- **Direct feeder tests** (`clients/web/turn-derivation.test.mjs`, 18 checks): repeated renders of one sample, 359/1° wrap both directions, duplicate/reorder, epoch reset, incarnation change, clock change, rapid reconnect, out-of-bound dt with recovery, missing-heading reset.
- **CI gap closed**: the video-identity and snapshot-association suites (merged in earlier PRs but never wired into CI) and the new turn-derivation suite now run in the workflow.

All nine node suites and the full Rust battery green locally.

luofang34:
Final review point addressed in aaf5646: a serially older sample now declares **nothing** — it neither advances state nor re-declares the cached rate, so a reordered input's age can never refresh the freshness of a rate it did not contribute to. Tests: `a serially older sample declares nothing at all` (reordered input with a fresh age → null) and `freshness after a reorder comes only from the accepted sample's age` (the next repeated render of the accepted sample re-declares with its own age, rate unchanged).

Issue #68, opened 2026-07-13
Harden isSnapshotIdentityValid(): enforce non-negative wire-range bounds on identity fields


luofang34:
### Context

Review note from the #67 merge (non-blocking): `isSnapshotIdentityValid()` currently accepts any integer-shaped `sourceId` / `sourceEpoch` / `sequence` without enforcing non-negativity or the corresponding wire integer ranges. Inputs today come only from accepted AV-01 ingress, so this is not an active defect — but a hostile-input / fault-injection pass must not find identity fields trusted beyond their wire domains.

### Required (when picked up)

- Bound every identity field to its wire type's range (non-negative; u8/u32/u64 as the AV-01 stamp defines) in `isSnapshotIdentityValid()` and any sibling validators on the video/association path.
- Reject out-of-range identities fail-closed with a typed reason, never clamp.
- Fault-injection tests: negative values, > wire-max values, non-integer numerics, BigInt/Number confusion at the u32/u64 seams.
- Sweep for the same pattern in the other browser-side validators (video identity, association, ingress mirrors) so the boundary rule is uniform.

SIM / NOT FOR FLIGHT.

luofang34:
Re-review of PR #73 at head `12c8177`: the two remaining hostile-input gaps are fixed.

- `SnapshotAssociator.associate()` now runs `metaFault()` before any BigInt mapping, addition, or comparison. Wrong-kind, negative, and over-range fields return a closed `malformed-meta` verdict with a typed fault; they no longer throw.
- `VideoIdentityTracker.admit()` returns the exact `{ field, rule }` fault and diagnostics retain `lastMalformedReason`, rather than exposing only the generic `malformed` bucket.
- The raw decoder surfaces over-u32 values without clamping, and the hostile tests' failure exit now runs after all added cases.

Verification: `wire-bounds`, wire decode, telemetry ingress, video identity, and snapshot association suites all pass locally; the adversarial association cases explicitly assert “never throws”; both GitHub CI runs are green; the PR is mergeable.

Verdict: implementation is merge-ready. Keep this issue open until PR #73 is actually merged into `main`, then close it as completed.

luofang34:
Completed by merged PR #73 (merge commit `d3802a5`) and re-verified on `main`.

The browser identity boundary now:
- enforces exact u8/u32/u64/i64 types and ranges across snapshot association, video identity, telemetry ingress, and raw decode;
[...]

PR #73, opened 2026-07-13
GEO-68: enforce exact unsigned wire ranges in browser identity validators

luofang34:
Implements #68. Hostile-input hardening of every browser-side snapshot, video, association, and ingress identity validator. **SIM / NOT FOR FLIGHT.**

## The gap

The identity validators accepted values outside their exact wire ranges: `snapshot-association.isSnapshotIdentityValid` took a source id as *either* a BigInt *or* a truncating Number and left epoch/sequence/acquisition-time unbounded; `video-identity.isMetaValid` left the camera/calibration ids and the u64 capture time unbounded; `telemetry-ingress.isStampValid` had no u64 upper bound (a decoded varint past 2^64 slipped through). A negative, fractional, over-range, or wrong-numeric-kind value could be admitted.

## The fix

- **One shared `wire-bounds.js`** with exact-unsigned-range predicates — `isU8`, `isU32`, `isU64`, `isIncarnation` — each rejecting negative, fractional, over-range, and **wrong-numeric-kind** values (a Number where a u64 BigInt is required, or a BigInt where a u32/u8 Number is required), fail-closed, never clamped.
- **Applied uniformly**: snapshot-association (source id = u64 BigInt strictly, epoch/sequence = u32, acquisition time = u64), video-identity (frame source id = u8, camera/calibration = u32, capture time = u64), telemetry-ingress (source id / acquisition time = u64 with the missing upper bound — reorder/freshness gating unchanged).
- **Hostile boundary tests** through the public `observe`/`admit` APIs: each width's negative, overflow (max+1), fractional, and Number/BigInt-confused case is refused (snapshot → `invalid` counter, never observed; video → `ADMIT.MALFORMED`). Plus an exhaustive `wire-bounds.test.mjs` unit suite, now registered in CI.

| Criterion (#68) | Evidence |
|---|---|
| Exact unsigned ranges across snapshot/video/association/ingress validators | shared `wire-bounds.js` applied in all three validators |
| Reject negative / overflowing / fractional / Number-BigInt-confused; never clamp | `wire-bounds.test.mjs` (per width) + hostile suites in snapshot-association / video-identity tests |
| Typed fail-closed reasons | snapshot `invalid` counter (never observed); video `ADMIT.MALFORMED` |
| Hostile boundary tests + full CI | new suites, `wire-bounds.test.mjs` wired into ci.yml |

## Note on ci.yml

This lane adds one browser-suite line (`wire-bounds.test.mjs`) to `.github/workflows/ci.yml` in the browser-suites section. The concurrent SVS-02 lane registers its new Rust crate in a different section; the two additions are non-adjacent and merge cleanly in either order.

## Verification

All nine (now ten) node suites green; the tightening rejects only out-of-range values that valid data never produces, so the existing telemetry-ingress / video-identity / snapshot-association / instruments / wire suites stay green. No Rust touched.

Refs #68.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,9 @@ jobs:
       - name: browser turn-derivation tests (DYN-01)
         run: node clients/web/turn-derivation.test.mjs
 
+      - name: browser wire-range identity bounds (GEO-68)
+        run: node clients/web/wire-bounds.test.mjs
+
       - name: Setup buf
         if: hashFiles('schemas/**/*.proto') != ''
         uses: bufbuild/buf-action@v1
--- a/clients/web/calibration.test.mjs
+++ b/clients/web/calibration.test.mjs
@@ -43,6 +43,8 @@ function fpvFrameMeta(overrides = {}) {
     mappingTargetClock: CLOCK_VEHICLE_BOOT,
     mappingOffsetNanos: 0n,
     clockErrorBoundNanos: 0n,
+    receiveTimeNanos: 0n,
+    publicationTimeNanos: 0n,
     ...overrides,
   };
 }
--- a/clients/web/snapshot-association.js
+++ b/clients/web/snapshot-association.js
@@ -23,7 +23,7 @@
 // match, published/recognized calibration, overflow-safe mapped time, error
 // budget) can never be bypassed.
 
-import { conformalGate, mapCaptureTime, DEFAULT_MAX_CLOCK_ERROR_NANOS } from "./video-identity.js";
+import { conformalGate, mapCaptureTime, metaFault, DEFAULT_MAX_CLOCK_ERROR_NANOS } from "./video-identity.js";
 import { firstFault } from "./wire-bounds.js";
 
 const CLOCK_VEHICLE_BOOT = 1;
@@ -47,6 +47,7 @@ export const ASSOCIATION = Object.freeze({
   STREAM_DISCONTINUITY: "stream-discontinuity",
   TOTAL_ERROR_EXCEEDS_BUDGET: "total-error-exceeds-budget",
   NOT_ADMITTED: "not-admitted",
+  MALFORMED_META: "malformed-meta",
 });
 
 // The first field of an AV-01 snapshot identity to violate its exact wire type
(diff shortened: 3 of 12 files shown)


luofang34:
Review addressed in b9c0741 — all five points, each with a test that now actually gates (verified: removing a guard fails the suite).

**1. Raw decoder no longer clamps.** `decodeMeasurementStamp` dropped the `>>> 0` on `sourceEpoch`/`sequence`, so a wire value past u32 is surfaced raw and refused by the identity validator fail-closed — not silently truncated to an in-range zero. `wire.test.mjs`: `a source_epoch past u32 is surfaced raw, not clamped to 0` + `the exact u32 max round-trips through decode unchanged`; `telemetry-ingress.test.mjs` proves the same value is refused end-to-end (`a source_epoch past u32 is refused, never clamped into range`) and the exact u32 max is accepted.

**2. Hostile tests now gate.** The suites' `if (failures > 0) process.exit(1)` was in the middle of the file with the hostile tests appended after it, so their failures never set the exit code. The exit check now runs LAST in every suite. Sanity-checked: removing the conformalGate guard makes `video-identity.test.mjs` exit 1.

**3. The associate() bypass is closed.** `conformalGate` now runs `isMetaValid(meta)` before any ready verdict, so the direct `associate()` path (which does not call the tracker's `admit()`) can no longer bless a malformed frame. Test: `a malformed meta (u8 sourceId=256) is refused by the gate, never ready` (+ a BigInt calibration id).

**4. Typed reasons are real.** `wire-bounds.js` gained `fieldFault`/`firstFault` returning a typed `{ field, rule }` (`wrong-numeric-kind` / `negative` / `fractional` / `out-of-range` / `malformed`). The snapshot associator records `lastInvalidReason` and the ingress gate records `lastRejectReason`, each naming the exact field and rule — the comment's claim is now backed by code. Tests assert `sourceEpoch/out-of-range`, `sourceId/wrong-numeric-kind`, `sourceIncarnation/malformed`.

**5. Raw wire→decode→ingress coverage.** New tests cover the over-u32 varint (refused, not clamped), the wrong-numeric-kind sourceId (a Number where a u64 BigInt is required → refused), and the exact-max boundary (accepted).

The shared `isU8/isU32/isU64/isIncarnation` design and the Number/BigInt boundary — which the review kept — are unchanged; the fixes are the raw-decode reject, the gate guard, the test-exit position, and the typed-reason implementation. All ten node suites green; committed only the ten clients/web files (no cross-lane bleed).

Issue #77, opened 2026-07-13
ASSURE-01: add a standard-neutral lifecycle evidence graph and scoped no-orphan gate


luofang34:
### Objective

Create one standard-neutral lifecycle evidence graph for Pilotage and a scoped no-orphan gate. The graph is the maintained engineering record; DO-178C, ISO 26262, ECSS, and future program views are projections over it rather than parallel trace databases.

### Architecture

- Model typed nodes for intended functions, hazards, safety goals/requirements, derived requirements, architecture/design, implementation units, verification cases/procedures/results, coverage analyses, reviews/approvals, configuration items, tools, anomalies, and external evidence.
- Model typed relations such as `derives_from`, `mitigates`, `allocated_to`, `satisfies`, `implemented_by`, `verified_by`, `result_of`, `covers`, `reviews`, `approves`, `justified_by`, `decomposes_to`, and `independent_from`.
- Keep stable IDs and content digests. Generated matrices and reports are views, not the source of truth.
- Keep assurance claims and standard objective mappings outside the core graph. Do not force the graph into a fixed `SYS -> HLR -> LLR -> TEST` hierarchy.

### Scoped no-orphan policy

For the declared in-scope assurance case:

- every safety-relevant requirement has an upstream intended-function/hazard/rationale path and a downstream design, implementation, verification-case, and verification-result path;
- every claimed result resolves to the executed case, configuration baseline, tool identity, and covered requirement;
- derived requirements have recorded safety impact, rationale, upstream notification/disposition, and review;
- exceptions are explicit records with owner, rationale, expiry/status, and independent review where required; absence or an empty graph is never reported as valid.

Different node kinds may have different permitted roots/leaves. The gate therefore enforces a declared policy and scope, not the false rule that every file or every non-safety test must attach to a certification objective.

### First vertical slice

Use #17 ATT-01 extreme-attitude behavior as the first end-to-end slice: intended function and display hazard -> behavioral requirement -> horizon design -> implementation -> raster cases at +/-89/90/91 degrees and inverted attitudes -> recorded outcomes -> review. Do not claim DO-178C or ASIL compliance from this slice.

### Acceptance criteria

- [ ] Versioned, documented, standard-neutral graph schema with deterministic serialization and validation.
- [ ] ATT-01 vertical slice resolves bidirectionally from behavior to current result and from result to behavior/configuration.
[...]

PR #86, opened 2026-07-14
ASSURE-01: standard-neutral lifecycle evidence graph and scoped no-orphan gate (#77)

luofang34:
ASSURE-01: standard-neutral lifecycle evidence graph and scoped no-orphan gate (#77)

## What this is

`pilotage-evidence` is one maintained engineering trace record — typed lifecycle
nodes and relations, stable ids, SHA-256 content digests, and a canonical,
byte-reproducible serialization — plus a scoped no-orphan gate that holds a
declared policy over a declared scope and **fails closed**. Program views
(DO-178C / ISO 26262 / ECSS) are projections held *outside* the graph.

**SIM / NOT FOR FLIGHT.** A passing gate establishes no DO-178C, ISO 26262,
ECSS, ARP4754A/4761, TSO, ASIL, certification, or operational credit, and is not
tool qualification (DO-330).

## ATT-01 vertical slice — honest verdict

`docs/instruments/evidence-graph.evg` models the extreme-attitude behavior slice
end to end and resolves both ways (`--trace`). Its review is genuinely
**pending**, so the gate reports a `review-incomplete` finding and a verdict that
is **not VALID** — "structurally traced, review pending" — never green-washing an
unfinished review.

## Evidence is resolved and matched, not asserted

- **Verification cases** name a real test symbol; resolution confirms each is
  defined in the referenced file.
- **Results** name a committed structured run-record artifact and pin it with an
  `output-digest`. Resolution reads the artifact, hashes it (a digest can never
  be detached hex), **and** parses its `command`, `config-digest`,
  `tool-version`, and `outcome` and requires each to match the result node — hash
  equality is not evidence the declared run happened. A source file (no run-record
  fields), a field mismatch, a missing artifact, or a hash mismatch all fail. The
  ATT-01 results reference real captured-output records under
  `docs/instruments/evidence-artifacts/att01/`.
- **Reviews** name a specific record entry by anchor (`path#anchor`). The gate
  resolves that one entry and verifies its reviewer, date, closed disposition, and
  coverage of every reviewed node — never a file-wide grep. A named entry that is
  missing or pending fails even when the file holds other complete entries.
- **Exceptions** expire: the CLI enforces an as-of date, so an expired exception
  cannot suppress a finding.

## Fail-closed: seventeen negative fixtures

Structural (fail plain validation): empty, missing-root, missing-required-layer,
dangling-edge, unresolved-selector, orphan-requirement, missing-result,
placeholder-result, pending-review, unbacked-review, unexecuted-result, and an
expired exception. Resolution-only (structurally valid, fail under
`--resolve-selectors`): unresolvable-selector-fs, unresolvable-artifact (no
committed artifact), mismatched-artifact (content ≠ digest), field-mismatch-
artifact (hashes correctly but the parsed command disagrees), source-as-artifact
(a source file with no run-record fields), and unresolved-review-entry (the named
entry is pending while a sibling is complete). Each has a test.

## CI stays advisory

The trace gate runs as a **separate, non-blocking** workflow, not part of the
required `quality-gates` job. It runs the structural fixtures, the
resolution-only fixtures under `--resolve-selectors`, and the expiry proof.

Refs #77.


luofang34:
Pushed three commits addressing the review findings:

- **eea1da3** — the `pilotage-evidence` Cargo.lock delta (branch passes `cargo check --locked`).
- **fcbd31c** — the structured run record now carries `run-id` and the gate cross-checks it against the result node (`result_artifact_fields` += run-id), so hash equality alone cannot bind evidence recorded against some other run. Every declared path (execution-output artifact, review record, case locator) now resolves through a contained resolver: absolute paths, parent ('..') components, and symlinks whose canonical target leaves the evidence root are refused before any filesystem read. Negative fixtures + tests cover the run-id mismatch and each escape class (the absolute-path fixture verified to actually read /etc/hosts on the previous code); artifact digests re-pinned for the run-id line. Artifact-resolution negatives moved to `tests/artifact_resolution.rs` to keep files under the 500-line gate.
- **d596498** — the four new fixtures join the evidence-gate workflow's must-fail loop.

All four fixtures verified to fail against the previous head. Full local CI + `cargo check --locked` green in an isolated worktree; the gate binary run against the real graph reports exactly one open finding — the honestly pending REVIEW-AIR01 — so the infrastructure can merge while #77/#17 stay open pending the real review disposition.

luofang34:
Pushed 7b979c5 addressing both remaining traceability gaps:

1. **Test-source binding** — the policy now requires `source-digest` (`git-blob:<sha1>`) on every result, and a new resolver hashes every `result-of` case locator under the contained root and requires equality. A result recorded against an older implementation of the same-named test fails closed (`stale-source-artifact.evg`; detail says the recorded result predates the test source). The three real-graph source-digests were verified to match the current tree.
2. **Genuine test identity** — `defines` no longer accepts a textual `fn name(`: the definition must be an actual function-definition line (comments and string literals are stripped, prefix limited to fn qualifiers) preceded by a `#[test]`-family attribute (path's final segment starts with `test`, covering `#[test]`, `#[tokio::test]`, `#[test_case(...)]`). `sample_tests.rs` now carries real attributes plus comment/string/helper decoys, and `selector-not-a-test.evg` pins the helper rejection. Unit tests cover all decoy classes.

Both new fixtures verified to fail against the previous head (d596498). Both joined the evidence-gate workflow's must-fail loop.

**Independent review recorded**: `rec-att-01` (reviewer sokoly — project owner, independent of the agent author; 2026-07-15; APPROVED, with rationale citing the defects found and verified across these review rounds), and REVIEW-AIR01 completed accordingly. The fully resolved gate on the real graph now returns **VALID**, with green-washing inverses (stripped record fields, pending status, unresolvable entry anchor) still failing closed. The AIR-01 baseline reviews (project-owner/safety/human-factors sections) remain PENDING and tracked separately per the roadmap.

Full local CI + `cargo check --locked` green in an isolated worktree; `sha1` added as a direct-version dep (git object id, not a security boundary) with only its own lockfile hunks committed.

Issue #20, opened 2026-07-11
SRC-01: add per-function source comparison, miscompare, and deterministic reversion


luofang34:
### Problem

AircraftState currently provides one value per estimate group and one global quality. It cannot represent independent candidate sources, compare them coherently, show disagreement, select/revert deterministically, or identify the displayed source. Two-source disagreement also cannot establish which source is correct.

### Scope

- Accept multiple candidates per display function with source ID, epoch, source/receive time, sequence, validity, integrity, accuracy, and datum.
- Compare only coherent samples within a configured skew and compatible units/datums.
- Model Agree, Miscompare, NotComparable, and InsufficientSources.
- Use function-specific comparison: shortest circular angle, q/-q-invariant quaternion angle, and explicit scalar/vector units.
- Put thresholds, persistence, hysteresis, priority, manual selection, reversion, and return policy in AirframeSourcePolicy.
- Provide deterministic states for primary selected, miscompare, reverted secondary, and all failed.
- Keep two-source disagreement ambiguous unless independent integrity evidence justifies selection; never silently average.
- Reset persistence on epoch change and prevent reversion chatter.
- Propagate selected source/reversion state to displays and typed transitions to ALR-01.
- Keep logic pure, bounded, and allocation-free.

### Non-goals

- Claiming physical independence or common-cause mitigation.
- Triple-modular voting or sensor fusion.
- Picking the closest-looking source.
- Certification claims.

### Acceptance criteria

- [ ] Displayed values always identify their selected source.
- [ ] Transient differences do not reversion; sustained differences produce explicit ambiguity.
- [ ] Failed primary can revert only according to policy and is visibly annunciated.
- [ ] No compatible source produces a failed state.
- [ ] Stale, reordered, old-epoch, skewed, or datum-incompatible data are never compared as simultaneous valid samples.
- [ ] Return-to-primary cannot chatter.

### Deterministic verification

Test agreeing sources, transient/sustained thresholds, two-source ambiguity, every source-failure combination, stale/skewed/restarted/reordered input, 359/1 heading, q versus -q attitude, incompatible altitude datums, hysteresis jitter, manual selection, and wrapping generation counters. No sleeps.

### Dependencies

AV-01, VAL-01, ALT-01, and NAV-01. Comparator API may be designed in parallel; ALR-01 independently consumes later transitions.

### Design inputs

[...]

PR #87, opened 2026-07-14
SRC-01: per-function source comparison, ambiguity, selection, deterministic reversion (#20)

luofang34:
Implements #20 (SRC-01). New `source_compare` domain module in `pilotage-instrument-state` (no_std, allocation-free, pure; time enters as caller-supplied ms).

- **Four states**: `Agree`, `Miscompare`, `NotComparable`, `InsufficientSources` (fail-closed default).
- **Function-specific comparison** via a `Comparable` trait: shortest circular angle (heading; magnetic/true never mixed without variation), q/−q-invariant geodesic quaternion angle (attitude), explicit scalar/vector units, datum-tagged altitude (cross-datum → NotComparable, never coerced).
- **AirframeSourcePolicy** (validated): priority, agree/miscompare hysteresis, skew budget, max age, sustain, return-stable, reversion/manual toggles.
- Two-source disagreement stays **ambiguous** unless a uniquely higher-integrity source justifies selection — never averages, never picks closest. Reversion only per policy; return-to-primary anti-chatter; persistence resets on epoch change.
- Propagates selected source + typed transitions to **ALR-01** (reuses `pilotage_alerts::MiscompareFault`; no parallel fault type). `wrapping_add(1)` generation.

Deterministic verification covers every issue item (agreeing, transient/sustained, two-source ambiguity, every failure combination, stale/skewed/restarted/reordered, 359/1 heading, q vs −q, incompatible datums, hysteresis jitter, manual selection, wrapping counters). Independently verified: reuses MiscompareFault, ambiguity + q/−q tests present, CI green on the rebased base. Draft for review.


luofang34:
Pushed 0eaa411 addressing the airspeed-unit review finding: scalar/vector differences are now expressed in the metric's canonical unit (m, m/s) before the policy thresholds apply — `ScalarUnit::canonical_factor` converts knots at exactly 1852/3600, and an unknown/mismatched unit yields an infinite difference (never agreement). Regression tests verified to fail against 67b18b2: an 8 kt split (~4.1 m/s, inside the hysteresis band) no longer annunciates a miscompare, while a 12 kt split (~6.2 m/s) still does; metric-level tests pin the canonical conversion for both `ScalarMeasure` and `VectorMeasure`. Cargo.lock now carries the branch's `pilotage-alerts` dependency — `cargo check --locked` passes on this branch. Full local CI (fmt, clippy -D warnings, all-target tests, doc gates, release build, thumbv7em no_std check, check-structure) green in an isolated worktree.

luofang34:
Pushed a60cf3a fixing the sequence-rollover counterexample: the usability gate now orders sequences modularly — `current.wrapping_sub(previous)` must be non-zero and strictly inside the forward half range — so `u32::MAX → 0` is one normal advance (producers count with `wrapping_add(1)`), duplicates and backward reorders are still dropped, and the ambiguous exact half-range step (0x8000_0000) is rejected. Four regression tests cover wrap acceptance, duplicate rejection, backward reorder rejection (including that the high-water mark does not regress), and both sides of the half-range boundary; the wrap and half-range tests verified to fail against the previous `<=` check. Full local CI + `cargo check --locked` green in an isolated worktree of this branch.

Issue #30, opened 2026-07-11
REN-04: gate backend conformance, goldens, fault injection, and resource budgets


luofang34:
### Objective

Build the renderer verification wall: backend semantic conformance, reference hashes, visual tolerances, fault injection, deterministic resource/work budgets, and target-bound timing evidence.

### Completed engineering baseline

- [x] Exact reference-raster hashes and a reviewed versioned golden corpus.
- [x] Documented semantic conformance for the simulator-only Canvas backend.
- [x] Deterministic tests for malformed/unknown/truncated scenes, backend failure, unsupported opcodes, resource exhaustion, liveness/recovery, and stale-image containment.
- [x] Hard CI gates for command count, stack depth, scene/framebuffer size, coverage/composite work, and per-cost-class render work.
- [x] Golden changes require explicit review metadata and are never rewritten automatically by CI.
- [x] A provisional per-class cost envelope with explicit assumptions and a display-derived SIM liveness deadline; this is not WCET or display-suitability evidence.
- [x] PR #93 placed the per-cost-class counters and provisional envelope on main.
- [x] PR #94 recorded a connected **PX4-profile** FC with raw capture/USB evidence while keeping MCU identity explicitly unverified.

### Remaining closure gates

- [ ] Maintain explicit hardware tiers: the available board may serve as a class high-end reference; select and measure a representative lower-tier render-target floor later.
- [ ] Detect and identify each running target over USB CDC/USB descriptors rather than relying on a generic name or asking for identity manually.
- [ ] Obtain direct MCU identity for any hardware-specific claim; the present artifact does not prove STM32H743.
- [ ] Run a dedicated render-measurement firmware/harness and commit an artifact binding firmware/build hash, detected MCU, measured clock, compiler/flags, cache/flash/memory configuration, target/features, tool identity, and raw output.
- [ ] Replace provisional cost estimates with target-grounded bounds and apply the real selected-display refresh/latency deadline.
- [ ] Gate the selected target in CI or a controlled hardware evidence run; exceeding the requirement fails closed.
- [ ] If full redraw cannot meet the requirement, record and implement the partial-redraw or acceleration decision before closure.

### Scheduling and claims

#30 remains open. The missing lower-tier board and measurement harness block target-performance/WCET claims and final M0 closure, not unrelated SVS, Evidence, HUD-SIM, vehicle-link, or video-policy work.

[...]

PR #93, opened 2026-07-15
REN-04: price counted render work into a provisional cost envelope (#30)

luofang34:
Begins the target timing work for #30 (**refs #30, does not close it** — #30 stays open pending a real USB CDC target identity, a bound measurement artifact, a real display deadline, and a possible hardware-acceleration decision).

## What changed

- **Per-class counted work** — `RenderWork` counts each cost class separately: integer polygon winding edge tests, f32 stroke segment-distance tests, circle/arc disc tests, **arc angular extras** (two cap `sqrtf`s, `atan2f`, `fmodf` — their own class, so an arc sample is never billed as a bare disc test), and source-over composites. Counting is worst-case per sample (early exits never under-count) and stays a pure function of scene bytes and dimensions — the gate cannot flake. Measured fixture split (PFD @480×360): 1,780,493 polygon-edge / 115,506 stroke-segment / 84,681 disc / 84,681 arc / 433,729 composites; each class budget is 2× the worse of the PFD/HSI fixtures.
- **Provisional cost envelope, not WCET** — `timing::TargetTimingModel` prices each class with its own conservative bound, stated as a zero-wait instruction estimate **doubled** as a memory-system allowance (polygon edge 40, stroke segment 80, disc 60, arc 240, composite 80 cycles), plus a fixed frame overhead. The API says what it is: `envelope_cycles` / `envelope_us` / `within_deadline`. The 480 MHz clock is recorded as an **assumption** the envelope scales against, not a fact. A WCET claim requires measured provenance.
- **Typed measurement binding** — `CycleProvenance::MeasuredUsbCdc` is a struct requiring the firmware identity + build hash reported over the CDC handshake, the MCU part, the measured clock, the compiler and flags, the cache/flash/memory state, and the committed raw measurement output path. A measurement that cannot state its configuration is not a measurement.
- **Deadline derived from a requirement, never from the envelope** — the frame deadline is the pre-existing SIM display liveness requirement (`PanelHealth livenessDeadlineMs = 1000 ms`, `clients/web/instrument-health.js`): a panel that cannot produce a frame inside it fails LIVENESS. The envelope prices at 605,209 µs against that 1,000,000 µs deadline — non-circular, with the margin consumed as the deadline tightens to a selected display's refresh requirement.
- **Detection, not asking** — `scripts/detect-target.sh` scans CDC device names and attempts an identity handshake (raw 115200, 2 s banner read); an unidentified device is reported as unable to ground a measured model. No target is currently connected, and the artifact records that.
- **Artifact + drift guard + CI gate** — `docs/instruments/evidence-artifacts/timing/target-timing.txt` records the detection outcome, per-class bounds with rationale, assumptions, budgets, the derived envelope, and the deadline source; `the_timing_artifact_matches_the_shipped_model` fails CI on any drift, and the `target timing gate (REN-04)` CI step runs the detection plus the envelope gate (`budget_envelope_fits_the_display_derived_deadline`).

## Verification

Isolated-worktree full CI: `cargo check --locked`, fmt, clippy `-D warnings`, all test suites 0 failures, doc gates, `check-structure.sh`, thumbv7em no_std check (the timing module is `no_std`), release build. Conformance goldens unaffected — the counters change no painted bytes. A regression test pins that an arc test is priced strictly dearer than a disc test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,10 +98,12 @@ jobs:
 
       - name: target timing gate (REN-04)
         # No display hardware is selected: the USB CDC scan documents that no
-        # target is connected, and the WCET gate prices the counted work
-        # budget with the conservative model against the recorded frame
-        # deadline. The artifact drift guard keeps the committed timing
-        # record equal to the shipped model.
+        # target is connected, and the envelope gate prices the per-class
+        # work budget with the conservative model against the frame deadline
+        # derived from the SIM display liveness requirement. The artifact
+        # drift guard keeps the committed timing record equal to the shipped
+        # model. The envelope is provisional — not a WCET claim — until a
+        # measured target binds firmware, clock, memory, and compiler state.
         run: |
           bash scripts/detect-target.sh
           cargo test -p pilotage-instrument-raster timing
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,16 @@ jobs:
           -p pilotage-instrument-raster
           --target thumbv7em-none-eabihf
 
+      - name: target timing gate (REN-04)
+        # No display hardware is selected: the USB CDC scan documents that no
+        # target is connected, and the WCET gate prices the counted work
+        # budget with the conservative model against the recorded frame
+        # deadline. The artifact drift guard keeps the committed timing
+        # record equal to the shipped model.
+        run: |
+          bash scripts/detect-target.sh
+          cargo test -p pilotage-instrument-raster timing
+
       - name: geospatial contract (SVS-01)
         # Datum discipline, coherence, view/projection, availability, and the
         # fail-closed versioned ABI for synthetic vision and conformal HUD.
--- a/crates/pilotage-instrument-raster/src/curve.rs
+++ b/crates/pilotage-instrument-raster/src/curve.rs
@@ -83,6 +83,8 @@ fn paint(
         let dy = (py as f32 + 0.5) - disc.cy;
         for px in region.left..region.right {
             surface.count_sample();
+            // One disc-distance test per sample, priced like an edge test.
+            surface.count_edge_tests(1);
             let dx = (px as f32 + 0.5) - disc.cx;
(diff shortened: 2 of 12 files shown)


luofang34:
Pushed f068540 addressing all four review findings:

1. **Cost classes split** — `RenderWork` now counts polygon edge tests, stroke segment tests, disc tests, **arc angular extras** (two cap `sqrtf`s, `atan2f`, `fmodf` — their own class, priced 4× a bare disc test, so the arc under-count is gone), and composites, each with its own conservative bound. Measured fixture split (PFD @480×360): 1,780,493 polygon edge / 115,506 stroke segment / 84,681 disc / 84,681 arc / 433,729 composites; class budgets are 2× the worse fixture. A regression test pins that an arc test is priced strictly dearer than a disc test.
2. **Provisional cost envelope, not WCET** — the API is now `envelope_cycles/envelope_us/within_deadline`; every bound is stated as a zero-wait instruction estimate **doubled** as a memory-system allowance, and the 480 MHz clock is explicitly recorded as an assumption the envelope scales against — not a fact. The docs and artifact say WCET applies only to a measured model.
3. **Measurement binding** — `CycleProvenance::MeasuredUsbCdc` is now a struct requiring firmware identity + build hash, MCU part, measured clock, compiler + flags, cache/flash/memory state, and the committed raw output path. `detect-target.sh` attempts a real CDC identity handshake (raw 115200, 2 s banner read) and reports an unidentified device as unable to ground a measurement.
4. **Deadline from a requirement, not the envelope** — the frame deadline is the pre-existing SIM display liveness requirement (`PanelHealth livenessDeadlineMs = 1000 ms`, `clients/web/instrument-health.js`): a panel that cannot produce a frame inside it fails LIVENESS. The envelope prices at 605,209 µs against that 1,000,000 µs deadline — non-circular, and the margin is consumed as the deadline tightens to a real display refresh requirement after hardware selection.

#30 stays open. Full local CI + `cargo check --locked` green in an isolated worktree (thumbv7em no_std check included); the conformance goldens are unaffected.

Issue #95, opened 2026-07-15
[TRACKING] VID-01: unified video encoding and decode strategy


luofang34:
## Role

Tracking epic for the video encode/decode architecture. Do not close this issue from one implementation PR; every child must remain independently revertible and carry its own regression guardrail.

Today the host emits MJPEG. The target architecture is one semantic strategy layer over platform backends and a software fallback, with explicit capability probing, negotiation, memory ownership, and failover.

## Active children

- [ ] #100 — browser H.264/WebCodecs decode lifecycle. Split this from PR #97 before merge.
- [ ] #102 — semantic encoder request/capability/negotiated-config types plus deterministic backend selection. This is the first encoder-side implementation and has no FFmpeg/vendor dependency.

After #102, file separate issues/PRs for the first H.264 NV12 low-latency software/libavcodec path, each hardware backend/probe, and each zero-copy memory adapter. HEVC/AV1 and Vulkan/native-vendor paths remain later work.

## Controlling architecture

1. Public requests use semantic parameters (`codec`, geometry/rate, usage, latency, rate control, quality, GOP, pixel format, hardware policy), never vendor option strings.
2. Capability enumeration is followed by a lightweight session probe. API presence, OS identity, or a driver's claim is not proof that a requested session can be created.
3. Backend choice is deterministic scoring over request fit, probe result, zero-copy path, copy/conversion cost, latency, quality, power, and known limitations. Hardware is not unconditionally first.
4. Frame storage and synchronization are typed (`Cpu`, `CvPixelBuffer`/IOSurface, DMA-BUF, Vulkan, CUDA or equivalent). Ownership and fences are part of the contract; GPU→CPU→GPU round trips are not treated as zero-copy.
5. The first common adapter may use libavcodec; native VideoToolbox/NVENC/etc. implementations are justified only by missing required capability, while implementing the same traits.
6. Negotiation returns the actual configuration and explicit compromises. `HardwarePolicy::Require` never silently selects software.
7. Interactive downlink defaults to H.264/NV12/8-bit, no B-frames/lookahead, short GOP, bounded VBV, periodic headers, force-IDR, and dynamic bitrate where proven. Recording is a separate encoder session.
8. Mid-stream backend/config changes require an explicit discontinuity/config event, decoder reset, and restart on an IDR. No plausible continuation across codec identity.
9. Browser decode uses WebCodecs; native/headless decode may use libavcodec. MJPEG remains a separate supported path.

## Candidate backend families

[...]

PR #97, opened 2026-07-15
REN: wasm-shared wire decode for the web viewer + H264 decode groundwork

luofang34:
## What & why

The demo viewer connected, took a control lease, and received video continuously, but **every video frame was dropped at the client identity gate with reason "(malformed)"**, and telemetry sat at "Missing (awaiting first sample)". This moves the viewer's real-time wire decode into the Rust wasm crate so it compiles from the **same wire definitions the host encodes with**, fixes the drift that caused the drops, and includes the host-side fix for the telemetry half.

## Root cause of the video drift — an exact field

It was a **validator** drift, not a byte-layout drift (the v2 header offsets in `wire.js` already matched `stream_tag.rs`). Aviate's video capture carries `CaptureClockMapping::Unavailable`, so the host writes `mapping_target_clock = 0` — the documented "absent target clock" sentinel. But `clients/web/video-identity.js` `metaFault` required `mappingTargetClock ∈ {1,2}` **unconditionally**, so it rejected every honestly-unavailable-mapping frame as `{field:"mappingTargetClock", rule:"malformed"}`. Confirmed against a live frame: `mappingAvailable=false`, `mappingTargetClock=0`, `metaFault` → malformed → dropped.

The fix binds the mapping's target clock to its availability (unavailable ⇒ `0`, present ⇒ a known clock), matching the encoder, and puts that contract in shared Rust so it can't drift again.

## Telemetry — the host-side half (also fixed here)

Telemetry sat at "Missing" because the **host emitted zero `TelemetrySample` datagrams**. Root cause (second commit, `adapters/aviate/src/shm.rs`): the aviate adapter opened `/aviate_gz_bridge` but rejected it on an exact `st_size == 216` check — macOS rounds a shm object up to the page size (16 KiB on Apple Silicon reports `st_size = 16384` for a 216-byte `ftruncate`), so the check always failed and the adapter silently fell back to MAVLink, where no telemetry flows. The fix: only a size *too small* to hold the state is a real mismatch (the mapping reads `SHM_SIZE` bytes regardless). With it, the adapter attaches to the shared-memory state block and broadcasts telemetry, so the PFD/HSI leave their fail-closed state.

## No-JS-decode architecture ("可以无需JS")

- **One definition of the v2 header**: `crates/pilotage-protocol/src/video_frame.rs` owns the offset table, `encode_v2`, `decode_v2`, and the capture-identity contract. `hosts/session-host` `frame_video_payload_v2` now *delegates its encode* to it, so host and browser read one layout.
- **wasm exports** in `clients/web-instruments` decode the two real-time surfaces from that shared code + the host's `prost` types:
  - `decodeVideoFrameV2(body)` → `{meta, fourcc, payloadOffset, payloadLen, fault}`
  - `decodeDatagramEnvelope(bytes)` → `{kind, message}` (the `TelemetrySample` shape, plus `Pong`/`FrameRejected`)
  - `u64`/`i64` cross to JS as **BigInt** and `u32` as **Number**, so the "wrong-numeric-kind" failure the JS gates fought by hand is impossible by construction.
- **JS keeps only transport plumbing and canvas paint.** `main.js` video and telemetry-datagram decode go through wasm; the stateful identity/conformal gates are fed the wasm-decoded meta. The one-time bootstrap handshake (ServerWelcome/LeaseResponse/AuthorityEvent) stays on the JS length-delimited reader — low-rate, stable, not a named drift surface.
- `video-identity.js` `metaFault` is fixed to match the contract and kept as the gate's own guard (defense in depth). `wire-wasm.test.mjs` loads the **real wasm** and pins that the wasm decode agrees with the JS reference on layout, numeric kinds, and the contract — including the unavailable-mapping case.

## H.264 decode groundwork (encode side = issue #95)

Today video is MJPEG only. This PR adds the **decode side** so a host that later emits H.264 renders without a viewer change: FourCC `"H264"` routes (in `main.js` `paintByCodec`) to a per-source WebCodecs `VideoDecoder` (`clients/web/video-h264.js`) — feature-detect, Annex-B NAL classification, keyframe-gated configuration (derives `avc1.PPCCLL` from the in-band SPS), and **fail-closed** with a typed log when WebCodecs or the profile is unavailable, behind the identity gate. Pure NAL/keyframe/codec-string helpers are unit-tested (`video-h264.test.mjs`).

The **encode side is tracked as issue #95** (unified semantic encoder layer → capability probing/scoring → libavcodec backends → x264 fallback; milestone M4). The wire additions this decode path assumes align to #95: FourCC `"H264"`, a keyframe flag, decoder-config delivery, and `StreamEvent::{CodecConfigChanged, Discontinuity, Keyframe}`. In-band Annex-B SPS on each keyframe is what the decoder configures against today, so a first encode-side cut can deliver config that way; the richer `StreamEvent` surface lands with #95.

## Follow-ups (not this PR)

- **Transport migration to `web-transport-wasm`** (confirmed direction): the client transport plumbing moves off the hand-written JS WebTransport calls to the `web-transport` crate family — in the browser build `web-transport-wasm`, using `ClientBuilder::with_server_certificate_hashes()` for cert-hash pinning; the host stays on wtransport 0.7.1 (protocol-compatible, untouched). That migration deletes the remaining hand-written transport JS, **including the Safari `datagrams.createWritable`/`writable` feature-detect in `startControlLoop`** (kept in this PR so Safari works until then — deliberately the only uncommitted hunk left on that file). Sequenced after this PR lands and is verified against the live stack; separate lane by size.
- **Native headless conformance client**: the same client code on `web-transport-quinn` can become a headless client that connects to the real wtransport host in CI (end-to-end wire conformance without a browser).
- Bootstrap-handshake envelope decode remains JS (stable, low-rate) — a clean follow-up to move to wasm with the same mechanism.

## Verification

Live against the running aviate/Gazebo stack: after the video fix, both the FPV and chase canvases paint continuously (100% of pixels), and frames are admitted with the honest log `video source N not conformal-ready: mapping-unavailable` — no more "malformed" drops. With the shm fix, the host attaches to the shared-memory state block and telemetry flows.

Local gates (blended tree **and** an isolated worktree of this branch = what CI sees): `cargo fmt --check`, `clippy --all-targets -D warnings`, `cargo test`, `cargo doc -D missing_docs`, `cargo build --release`, `scripts/check-structure.sh`, and the web node tests (wire-wasm 30, h264 8, video-identity 87, wire 42, telemetry-ingress 19) all pass. New CI steps run the wasm conformance and H.264 classification tests.


luofang34:
Review verdict: **do not merge the current head**, despite clean mergeability and green CI. It combines three independently revertible concerns and each needs its own issue/guardrail:

- #98 — shared Rust/wasm real-time wire decode.
- #99 — page-rounded Aviate shm capacity handling.
- #100 — browser H.264/WebCodecs lifecycle.

Concrete blockers found in the code:

1. `video_frame::decode_v2` converts `mapping_available` with `byte == 1`. A non-canonical raw value such as `2` becomes `false`; with target clock `0` it then passes `contract_fault()`. Reject raw values other than `0`/`1` before normalization and add the hostile Rust + real-wasm vector in #98.
2. The shm change has no regression test. It also converts a negative `st_size` to `u64::MAX`, which would be accepted, and comments call `actual >= 216` a layout check. Split #99, reject negative/short values, test negative/215/216/page-rounded capacities, and describe this only as a capacity check. Aviate#262 remains the real magic/version/layout gate.
3. `h264Decoders` survives transport replacement. `H264CanvasDecoder.configure()` captures the first session's `isActive` closure, so a reconnect reuses a decoder whose output callback tests a retired token and drops every new-session frame. Codec/incarnation state also crosses the session. Split #100 and make decoder ownership/reset explicit on session, incarnation, discontinuity, and config changes; current tests cover NAL classification only, not decoder lifecycle.
4. The uncommitted `datagrams.createWritable()` compatibility hunk is not part of this PR and is now tracked separately as #103. Do not append it here.

After splitting and fixing these points, #98 and #99 can merge without waiting for Aviate#262/#263, subject to fresh full CI. #100 should remain a separate child of #95.

Targeted review verification on the blended tree: `pilotage-protocol` 58/58, `pilotage-instruments-web` 17/17, `pilotage-adapter-aviate` 67/67, and the Node H.264 classification/video identity suites all passed. Those passes do not exercise the three counterexamples above.

luofang34:
Superseded by the split requested in the review: #98 (WEB-01, wire decode + non-canonical mapping_available fix), #99 (LINK-02, page-rounded shm capacity), #100 (VID-02, session-safe H.264 decoder, stacked on #98). Each is independently CI-clean in an isolated worktree; the createWritable hunk stays out (tracked as #103).

PR #104, opened 2026-07-15
LINK-02: accept a page-rounded Aviate shm capacity (#99)

### LINKED ISSUE #99: LINK-02: accept page-rounded Aviate shm capacity without claiming layout compatibility
## Objective

Restore Pilotage's read-only attachment to Aviate's current POSIX shared-memory block when the kernel reports page-rounded object capacity, without representing a capacity check as layout/version validation.

This is the narrow `adapters/aviate/src/shm.rs` change currently mixed into PR #97.

## Acceptance criteria

- [ ] Treat `st_size` as mapping capacity: negative or smaller-than-required values fail closed; exact and page-rounded capacities are accepted.
- [ ] Map/read only the current required byte count.
- [ ] Extract the capacity decision into a testable function with counterexamples for negative, one-byte-short, exact, and representative page-rounded values.
- [ ] Error names and comments say capacity/size, not layout compatibility.
- [ ] Keep the hand-written 216-byte layout and `OFF_*` constants explicitly provisional; no claim of structural compatibility follows from `actual >= required`.
- [ ] A live attach regression test or recorded reproducible host-side check confirms the current macOS behavior.
- [ ] Full repository CI is green.

## Dependency boundary

This fix does **not** wait for luofang34/Aviate#262. It restores compatibility with today's producer. Aviate#262 remains the gate for magic/version/declared-size validation, generated bindings, proper seqlock behavior, and deletion of Pilotage's offset mirror.

Silent shm-to-MAVLink downgrade/session visibility is separate LINK-03 work under #96.

SIM / NOT FOR FLIGHT.

luofang34:
Split out of #97 per the review verdict. Narrow `adapters/aviate/src/shm.rs` change (LINK-02).

## Problem
The aviate adapter opened `/aviate_gz_bridge`, then rejected it on an exact `st_size == 216` check. macOS page-rounds a shm object (a 216-byte `ftruncate` reports `st_size = 16384` on a 16 KiB page), so the check always failed and the adapter silently fell back to MAVLink, where no telemetry flows. The old code also coerced a negative `st_size` to `u64::MAX` (which would be accepted), and its comment called `actual >= 216` a layout check.

## Fix
- Extract a testable `admissible_capacity(st_size: i64, required) -> Option<u64>`: refuses a negative or too-small size, accepts an exact or page-rounded one, and maps only the required byte count.
- Rename the error to `ShmCapacityTooSmall { required, observed: i64 }` so a negative size reports honestly.
- Mark the 216-byte size and the `OFF_*` offsets explicitly provisional: `actual >= required` proves only "big enough to map", never that the fields still match. The magic/version/declared-size gate remains Aviate#262.

## Tests
`admissible_capacity` counterexamples: negative, one-byte-short (215), exact (216), and page-rounded (16384). Verified live against the running host — the fixed adapter attaches to the shm state block and telemetry flows.

Does not wait for Aviate#262. Closes #99.

### DIFF
--- a/adapters/aviate/src/error.rs
+++ b/adapters/aviate/src/error.rs
@@ -38,14 +38,17 @@ pub enum AviateAdapterError {
         #[source]
         source: std::io::Error,
     },
-    /// The shared-memory object's ABI size does not match the reader.
-    #[error("Aviate state shm {name} has {actual} bytes; expected {expected}")]
-    ShmSizeMismatch {
+    /// The shared-memory object is too small to back the reader's mapping.
+    /// This is a capacity check, not a layout/version check: it proves only
+    /// that at least `required` bytes are present, never that the block's
+    /// fields match (that gate is Aviate#262).
+    #[error("Aviate state shm {name} reports {observed} bytes; needs at least {required}")]
+    ShmCapacityTooSmall {
         /// The POSIX shm object name.
         name: String,
-        /// Required byte size.
-        expected: usize,
-        /// Observed byte size.
-        actual: u64,
+        /// Byte count the reader must be able to map.
+        required: usize,
+        /// The kernel-reported object size (`st_size`), which may be negative.
+        observed: i64,
     },
 }
--- a/adapters/aviate/src/error.rs
+++ b/adapters/aviate/src/error.rs
@@ -39,9 +39,9 @@ pub enum AviateAdapterError {
         source: std::io::Error,
     },
     /// The shared-memory object is too small to back the reader's mapping.
-    /// This is a capacity check, not a layout/version check: it proves only
-    /// that at least `required` bytes are present, never that the block's
-    /// fields match (that gate is Aviate#262).
+    /// This is a capacity check, not a layout/version check: capacity does
+    /// not establish layout compatibility — it proves only that at least
+    /// `required` bytes are present, never that the block's fields match.
     #[error("Aviate state shm {name} reports {observed} bytes; needs at least {required}")]
     ShmCapacityTooSmall {
         /// The POSIX shm object name.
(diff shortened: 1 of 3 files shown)


luofang34:
Recorded reproducible attach check (per review + #99), from the live verification session on this machine:

**Environment**: Darwin 27.0.0 arm64 (Apple Silicon MacBook Pro), `getconf PAGESIZE` = **16384**. Producer: Gazebo Harmonic headless (`gz sim -s -r --headless-rendering sim/worlds/x500_flightdeck.sdf`, `GZ_IP=127.0.0.1`, AviateGzPlugin creating `/aviate_gz_bridge` with `ftruncate(216)`), plus `sitl-gazebo-x500` (FC in Ready).

**Command**: `GZ_IP=127.0.0.1 RUST_LOG=info ./target/debug/pilotage-session-host --port <p> --adapter aviate`

**Key output, before the fix** (2026-07-15T16:44:58Z, pre-fix binary):
```
INFO pilotage_adapter_aviate::adapter: Aviate shm not available; using MAVLink/UDP error=Aviate state shm /aviate_gz_bridge has 16384 bytes; expected 216
```
i.e. the kernel page-rounded `st_size` (216 → 16384 on 16 KiB pages) and the exact-match check rejected a healthy block, silently degrading to the telemetry-less MAVLink path.

**Key output, after the fix** (2026-07-15T16:45:59Z, same producer stack, rebuilt binary):
```
INFO pilotage_adapter_aviate::adapter: Aviate link: shared-memory state block
```

**Block-health cross-check** (independent of the adapter, raw `shm_open`+`mmap` probe of 216 bytes): `seq` advanced 500 in 0.5 s (1 kHz physics tick), `valid=1` — the rejected object was serving live data throughout. Downstream confirmation: telemetry datagrams then flowed end-to-end (browser PFD/HSI rendering live pose).

Reproduce on Linux: `st_size` reports exactly 216 there (tmpfs does not page-round the reported size), so the capacity path admits it identically; the unit tests pin the negative/short/exact/page-rounded boundaries portably.

Issue #103, opened 2026-07-15
WEB-02: feature-detect the WebTransport datagram send stream


luofang34:
## Objective

Make the browser control loop acquire the WebTransport datagram send stream through the standardized API while retaining a bounded compatibility path for engines that still expose the older surface.

This tracks the current uncommitted `clients/web/main.js` hunk. It must not be smuggled into PR #97.

## Acceptance criteria

- [ ] Prefer `transport.datagrams.createWritable()` when it exists; otherwise use the legacy writable stream only when present.
- [ ] If neither API exists, surface one fail-visible control-unavailable state and do not claim a usable control lease.
- [ ] Tests cover standardized API, legacy API, neither API, writer acquisition failure, and session teardown without browser-version assumptions.
- [ ] Comments describe the feature-detection invariant, not which named engine is "current" or "old".
- [ ] Keep certificate pinning, backpressure, sample-time stamping, and session-token ownership unchanged.
- [ ] Record the eventual transport-crate migration as a separate issue; this compatibility fix remains independently revertible.
- [ ] Full repository CI is green.

SIM / NOT FOR FLIGHT.

PR #106, opened 2026-07-15
VID-02: session- and incarnation-safe browser H.264 WebCodecs decode (#100)

### LINKED ISSUE #100: VID-02: make browser H.264 WebCodecs decode session- and incarnation-safe
## Objective

Add a session-safe browser H.264 Annex-B decode path through WebCodecs behind the existing capture-identity gate. This is a child of tracking issue #95 and must be split out of PR #97.

## Acceptance criteria

- [ ] `H264` FourCC routes only admitted frames to a per-source decoder; MJPEG remains unchanged.
- [ ] Start only from an IDR/keyframe with usable in-band SPS/PPS or an explicit decoder configuration.
- [ ] Decoder state is owned by the active transport session and source incarnation. Session replacement, disconnect, discontinuity, or incarnation/config change closes and recreates it.
- [ ] No callback captures a retired session token and then governs a later session.
- [ ] Unsupported WebCodecs/profile/configuration and asynchronous decoder errors produce one typed fail-visible state, not an exception or permanent blank canvas.
- [ ] Tests exercise decoder lifecycle with an injected/fake `VideoDecoder`: initial keyframe, delta-before-keyframe, reconnect, source reset, configuration change, failure, and frame close.
- [ ] At least one browser or recorded real-stream check decodes a known H.264 Annex-B fixture.
- [ ] Full repository CI is green.

## Dependencies

This viewer work need not wait for the encoder implementation, but its accepted wire/config semantics must match #95. It must not claim end-to-end H.264 availability until an encoder-side child of #95 lands.

SIM / NOT FOR FLIGHT.

luofang34:
Adds H.264 (Annex-B) decode to the web viewer behind the ADR-0016 codec FourCC dispatch, with every decision in shared Rust and the decode proven in a real browser. Closes #100.

## What decides, and where

All byte- and state-level decisions live in `pilotage_protocol::h264` and reach the viewer through the wasm exports — there is no JS parser, validator, or ownership state machine:

- **Classification** (`h264.rs`): NAL iteration (3- and 4-byte start codes), keyframe determination, and the `avc1.PPCCLL` codec string from the SPS. Fail closed:
  - Parameter sets count only when they **precede** the IDR — a decoder consumes the stream sequentially, so `IDR → SPS → PPS` is `UndecodableKeyframe`, not decodable.
  - Bytes with no NAL units at all are `Invalid` (the session fails on them), never `Delta`.
  - The codec string is derived only from bytes inside the SPS unit itself; a truncated SPS never borrows its neighbor's bytes.
- **Decode session machine** (`h264/session.rs`): unconfigured → configured(codec) → failed/retired, returning one action per chunk (`configure-and-feed` / `feed` / `drop` / `fail`). A failed session surfaces one typed reason, then drops everything; retirement is permanent.
- **Decoder generations**: every (re)configuration, platform failure, and retirement advances a generation counter. A WebCodecs callback captures its generation at configure time and is honored only while `accepts_output_from(generation)` holds — so a callback retained by a **replaced, reset, failed, or retired** decoder can never paint over its successor or over the failure marker, including an in-band codec change within one session.
- **Per-source decoder ownership** (`SourceOwnership`): keyed by the transport session token's numeric generation; `reuse` / `build` / `replace` transitions keep a decoder from outliving its session or spanning a capture discontinuity.

`clients/web/video-h264.js` is the thin layer the platform forces: it executes returned actions against `VideoDecoder`, reports platform failures back into the machine, and blits frames. Unsupported codec/browser fails visible (typed log + canvas marker), never a silent blank.

## Proven in a real browser

`video-h264.browser.test.mjs` (gated in CI) launches headless Chromium against a harness page that decodes the recorded encoder fixture through the **actual viewer path** — wasm session machine + platform adapter:

- all five fixture frames output, each 48×32,
- every frame closed by the adapter before returning,
- decoded pixels actually painted on the canvas,
- no session failure, no failure log.

No Chromium available is a test **failure**, not a skip.

## Tests

- Rust (`h264/tests.rs`, 18 tests): NAL iteration, strict ordering counterexamples (`IDR → SPS → PPS`, no-NAL garbage, truncated SPS), fixture classification with pinned SHA-256, session machine transitions, ownership transitions.
- Wasm conformance (`wire-wasm.test.mjs`): the browser-side classifier agrees with the Rust tests on kinds, codec string, and typed reasons, over the same fixture.
- Adapter (`video-h264.test.mjs`, real wasm, fake `VideoDecoder`): lifecycle, failure surfaces, registry — and stale-callback tests that **invoke the retained old callbacks directly** after a codec change, after `close()`, and after a platform failure, asserting nothing paints, frames still close, and a late error cannot re-log or poison the successor.
- Browser boundary: the Chromium fixture decode above.

## Fixture

`crates/pilotage-protocol/tests/fixtures/h264-annexb-baseline.h264` — deterministic ffmpeg/libx264 `testsrc` encode (baseline, `avc1.42c00a`), provenance and digest in the adjacent `.provenance.md`, digest pinned in the Rust tests.


luofang34:
Rebased onto main (retargeted base to `main`; #105 is merged, so this now shows only the #100 delta) and addressed every point from the review. Not merging — ready for re-review.

1. **PPS check.** `video-h264.js` adds `hasParameterSets(bytes)` (SPS **and** PPS); `decode()` requires both in-band on a keyframe before configuring — an SPS-only keyframe now fails visible with a typed SPS/PPS reason instead of configuring a decoder that would stall.
2. **Real ownership tests.** The per-source ownership logic is extracted into a testable `H264DecoderRegistry` (injected decoder factory) that `main.js` now uses. Tests exercise the real transitions: same source+token reuses one decoder; a new token (session replacement mid-stream) closes the retired decoder and builds one bound to the new token; a discontinuity `reset(source)` closes only that source and rebuilds it on next use; `closeAll()` (teardown) empties the registry; and an end-to-end case flips the live token and asserts the retired-token decoder stops painting while the new one supersedes it.
3. **Failure paths actually driven.** New tests drive both: a `decodeThrows` fake makes `decode()` throw (sync path) → fail-visible; and the async path invokes the decoder's captured `error` callback → fail-visible + decoder closed. The claimed decode-failure coverage now exists.
4. **Real Annex-B fixture.** `clients/web/fixtures/h264-annexb-baseline.h264` is a 1640-byte libx264 Annex-B stream (`avc1.42c00a`; NALs 7,8,6,5,1…) with a provenance note (ffmpeg version + exact command over synthetic `testsrc`, no external media) and a pinned SHA-256. The test verifies the digest and runs the parser/keyframe/config-extraction path over the real bytes. WebCodecs decode itself needs a browser (CI is Node), documented in the header + provenance.
5. **Comments.** Removed "old/previous decoder" temporal wording in `video-h264.js` and `main.js`; the docstrings state the present mechanism (SPS+PPS required, held decoder replaced on codec change / session change).
6. **Rebase + gates.** Base retargeted to `main`; verified CI-clean in an isolated worktree of the branch: `check-structure`, `cargo fmt --check`, the wasm build, and all web node suites (video-h264 now 43 checks incl. the fixture). #106 has no Rust changes.

The `datagrams.createWritable` hunk (#103) stays out (uncommitted in the working tree).

luofang34:
Owner-directed restructure pushed (c919611): the pure byte logic — NAL iteration, keyframe determination, in-band SPS/PPS presence, avc1 codec-string derivation — moved out of hand-written JS into shared Rust (`pilotage_protocol::h264`: typed `ChunkClass`/`KeyframeFault`), reaching the browser through the wasm export `classifyH264Chunk` — the same drift-proof arrangement as the merged #105 wire decode, and directly reusable by a future native decoder (#95). The recorded libx264 Annex-B fixture moved into the protocol crate where Rust unit tests pin its provenance digest, full NAL sequence, and classification; the wasm conformance suite classifies the same bytes through the real wasm module. `video-h264.js` keeps only the WebCodecs session layer (decoder ownership, configure/feed, paint) with the classifier injectable; the JS suite now tests the session layer alone — both decode-failure paths and every registry ownership transition. Follow-up (with the web-transport-wasm migration): bind the WebCodecs session itself via web-sys, ending hand-maintained JS beyond generated glue and the boot entry. Not merged — awaiting owner re-review.

Issue #117, opened 2026-07-16
DEPLOY-01: one-line host provisioning and browser-only access with OIDC/passkey-only authentication


luofang34:
Target experience: a user provisions a Pilotage host on a server with one line (installer, script, or container) — the session host plus a chosen vehicle backend — and everything after that happens in a browser page. No local toolchain, no hand-assembled URLs.

## Scope to design

- **One-line install**: host + selected FC/simulator backend. The backend is adapter-pluggable: Aviate SITL/Gazebo today; JSBSim, X-Plane (#8/#9), AirSim, and other XIL backends later. The installer choice must not fork the host — same session wire, different adapter.
- **Browser-only client**: the host serves the viewer itself (the dev-only static server + ready-URL row already tracked in #11 is the seed of this); today's flow (separate static server, certificate hash copied by hand from the `LISTENING` line, ephemeral cert that dies with the process) does not survive a restart and cannot be given to a second user.
- **Authentication for anything reachable beyond loopback**: OIDC and/or passkeys (WebAuthn) ONLY. Password authentication is explicitly disallowed — no password form, no fallback. Authorization must compose with the existing session-authority semantics (#10) rather than bypass them; certificate pinning and join-token semantics stay intact (see #11 boundaries).
- Fits the LINK-01 (#96) target topology: browser → WAN → ground host (authority) → vehicle link.

Recording only — no work scheduled on this issue yet.

SIM / NOT FOR FLIGHT.

Issue #11, opened 2026-07-09
[TRACKING] DEMO-01: orchestration, headless conformance, and visual regression


luofang34:
## Role

Tracking issue for local demo orchestration and automated client verification. Implement each remaining row as a separate, independently revertible issue/PR when work starts.

## Status

- [x] Node instrument/scene/conformance tests run in CI, including `instruments.test.mjs` and the Canvas interpreter/scene corpus.
- [x] Session-host loopback tests exercise negotiation and datagram progress without visual browser inspection.
- [ ] Add a dev-only session-host static web server and print a ready URL with host/port/certificate parameters.
- [ ] Add repository orchestration recipes for wasm build, demo, local CI, and optional sibling Aviate launch without creating a source dependency between repositories.
- [ ] Add a native headless WebTransport conformance client against the real host; coordinate this with #10.
- [ ] Add fixed-state browser visual regression only as a backstop after semantic assertions; define tolerances and artifact review rather than manual pixel inspection.

## Boundaries

Automation must preserve certificate pinning and session authority semantics. Empty test join tokens are fixtures, not a production authentication claim. No first-party submodule is introduced.

SIM / NOT FOR FLIGHT.

luofang34:
Data point from the 2026-07-16 live SITL session reinforcing the two unchecked orchestration rows: bringing the demo up by hand takes four processes with per-process environment, in order —

1. `gz sim -s -r --headless-rendering sim/worlds/x500_flightdeck.sdf` with `GZ_SIM_SYSTEM_PLUGIN_PATH` (Aviate plugin build dir), `GZ_SIM_RESOURCE_PATH` (PX4 models), `GZ_IP=127.0.0.1`;
2. Aviate `sitl-gazebo-x500`;
3. `pilotage-session-host --port 4433 --adapter aviate` with `RUST_LOG`, `GZ_IP`, `PILOTAGE_AVIATE_LINK=shm`;
4. a static server for `clients/web` plus a viewer URL hand-assembled from the host's `LISTENING <port> <cert-hex>` line.

[...]

PR #125, opened 2026-07-17
DEMO-01: cargo xtask sim — one-command SITL session launcher with a backend seam

### LINKED ISSUE #124: DEMO-01: cargo xtask sim — one-command SITL session launcher with a backend seam
Implements the orchestration row of #11 ("repository orchestration recipes … optional sibling Aviate launch without creating a source dependency between repositories") as an in-repo `xtask` binary, per the runbook recorded in that issue's comments.

## Why xtask (not just)

The launcher is a supervisor, not an alias list: per-process environment assembly with platform differences, ordered startup with real readiness gates (gz topics → FC Ready → host `LISTENING` → viewer served), `LISTENING <port> <cert-hex>` parsing into a pinned viewer URL, clean teardown on ctrl-C or any child dying, and — later — other FC backends (PX4, JSBSim). That is logic, and bash-behind-`just` is where this codebase family's launch logic has already rotted twice (silent `set -e -o pipefail` death; the Linux `.so` hard-check that skips the plugin on macOS). An xtask is plain Rust under the same lint wall, testable, needs no extra tool beyond the toolchain, matches Aviate's existing xtask precedent, and its core can graduate into #117's one-line provisioning binary instead of being rewritten.

## Required

- `tools/xtask` crate + `.cargo/config.toml` alias so `cargo xtask …` works.
- `cargo xtask sim [--fc aviate] [--profile physical|simulation|oracle-only] [--port N] [--viewer-port N] [--open]`:
  - `SimBackend` trait seam; one `aviate-gz` implementation now (PX4/JSBSim explicitly deferred — the flag exists, unknown values fail typed).
  - Sibling Aviate located by `AVIATE_DIR` (default `../Aviate`), artifacts validated with actionable errors (plugin dylib, FC binary) — no source dependency.
  - Ordered launch with event-based readiness and deadlines, no sleeps-as-logic; logs per process under `target/`.
  - Parses the host's `LISTENING` line and prints/opens the viewer URL; certificate pinning semantics untouched (#11 boundary).
  - Refuses to start over stale session processes (clear pids + hint), tears down its own children (process groups, TERM then KILL) on ctrl-C or when any child exits.
- `cargo xtask reset` wrapping `scripts/reset-flight-sim.sh`.
- Guardrails in the same PR: unit tests for the `LISTENING` parser, URL construction, backend/profile selection fail-closed, environment assembly, and artifact validation.

## Out of scope

PX4/JSBSim implementations, `cargo xtask ci`, host `--serve-web` (separate #11 row; when it lands, the viewer-serving step here shrinks), #117 provisioning/auth.

SIM / NOT FOR FLIGHT.

luofang34:
Closes #124. Implements the orchestration row of #11.

## What this is

`cargo xtask sim` replaces the recorded four-process manual runbook (#11 comments) with one supervised command: it builds the session host, launches headless Gazebo with the Aviate plugin, the SITL FC, the host, and the static viewer **in order, each gated on an event-based readiness signal with a hard deadline** (gz world topics → FC `Ready` log line → host `LISTENING` line → viewer TCP accept), prints the pinned viewer URL (host port + certificate hash parsed from the `LISTENING` line — certificate-pinning semantics untouched, per #11's boundary), then supervises: ctrl-c or any stage dying tears everything down in reverse order via per-stage process groups (TERM, bounded grace, KILL) — no orphaned gz/FC/host processes, which this repo's sessions have repeatedly leaked by hand.

`cargo xtask reset` wraps `scripts/reset-flight-sim.sh`.

## Shape

- **Backend seam**: a `SimBackend` trait plans the simulator/FC stages and contributes the host adapter + environment; `--fc aviate` is the only implementation (PX4/JSBSim are future implementations of the same trait, per the issue — unknown names fail typed). Profile selection (`--profile physical|simulation|oracle-only`) mirrors the host's fail-closed vocabulary exactly.
- **No source dependency on Aviate**: the sibling checkout is located by `AVIATE_DIR` (default `../Aviate`) and every artifact is validated with an actionable hint (plugin dylib → the cmake line; FC binary → the cargo line). The macOS/Linux plugin-name difference is handled — the launcher does not inherit `launch_gazebo.sh`'s Linux-only `.so` check.
- **Refuses stale sessions**: running `gz sim`/FC/host processes are listed (via `ps`, not killed) with the exact cleanup command — the launcher only ever kills processes it started.
- Stage logs land under `target/xtask-sim/` and every readiness failure names the log it captured.
- `.cargo/config.toml` carries only the `cargo xtask` alias.

## Verified live (this machine)

- Stale-session refusal fired correctly against a real leftover session and listed it readably.
- Clean field: full launch → `session ready: http://127.0.0.1:8080/index.html?host=…&cert=…` (viewer answered 200), SIGINT → ordered reverse teardown, zero leftover processes.

## Guardrails

Unit tests: `LISTENING` parse (exact-format, rejects lookalikes/oversized ports/trailing tokens), viewer-URL pinning, backend and profile selection fail-closed, artifact validation hints, and the full planned gz/FC environment (plugin path, resource path, `GZ_IP`, headless `DISPLAY` removal) against a scaffolded checkout.

## Out of scope

PX4/JSBSim implementations, `cargo xtask ci`, host `--serve-web` (separate #11 row — it will retire the python viewer stage here), #117 provisioning/auth.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,4 @@
+# `cargo xtask <command>` — the repo's orchestration entry point
+# (tools/xtask): launches SITL sessions, wraps the reset script.
+[alias]
+xtask = "run -q -p xtask --"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2446,6 +2446,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "xtask"
+version = "0.1.0"
+dependencies = [
+ "thiserror",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "yasna"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ members = [
     "adapters/aviate",
     "tools/hid-probe",
     "tools/loopback-probe",
+    "tools/xtask",
     "hosts/session-host",
     "adapters/gazebo",
     "clients/web-instruments",
--- a/scripts/reset-flight-sim.sh
+++ b/scripts/reset-flight-sim.sh
@@ -17,6 +17,17 @@ gz service -s "/world/${WORLD}/control" \
 
 echo "restarting FC..."
 pkill -9 -f sitl-gazebo-x500 2>/dev/null || true
+
+# When `cargo xtask sim` supervises the session, the supervisor restarts
+# the flight controller itself; a script-spawned second FC would fight it
+# over the shm writer lease.
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+SUPERVISOR_PID_FILE="${REPO_ROOT}/target/xtask-sim/supervisor.pid"
+if [[ -f "${SUPERVISOR_PID_FILE}" ]] && kill -0 "$(cat "${SUPERVISOR_PID_FILE}")" 2>/dev/null; then
+  echo "done — the xtask supervisor restarts the FC; re-arm from the browser once it logs ready"
+  exit 0
+fi
+
 sleep 1
 AVIATE_DIR="${AVIATE_DIR:-$HOME/Aviate}"
 nohup "${AVIATE_DIR}/target/debug/sitl-gazebo-x500" > /tmp/fc_manual.log 2>&1 &
--- a/tools/xtask/Cargo.toml
+++ b/tools/xtask/Cargo.toml
@@ -10,6 +10,6 @@ workspace = true
 
 [dependencies]
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt", "signal", "time"] }
+tokio = { workspace = true, features = ["macros", "rt", "signal", "sync", "time"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
--- a/tools/xtask/Cargo.toml
+++ b/tools/xtask/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "xtask"
+description = "Repository orchestration: one-command SITL session launch and reset (DEMO-01)"
+version = "0.1.0"
+edition.workspace = true
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt", "signal", "time"] }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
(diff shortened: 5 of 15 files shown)


luofang34:
E2E test (browser, live SITL) surfaced one real defect, fixed in 5a53fed: the supervisor treated any stage death as fatal, so `cargo xtask reset` (and the browser Reset button — same script) killed the session it was resetting. The supervisor now restarts a dead flight-controller stage in place (bounded at 3; every other stage death stays fatal), and a `supervisor.pid` marker tells the reset script to skip its own FC respawn so two FCs never fight over the shm writer lease.

Full pass verified live: connect → arm → climb (truth d=−5.5 m) → land → disarm → `cargo xtask reset` mid-session (script deferred to supervisor, FC restarted 1/3, host reattached: shm epoch 2 + MAVLink clock epoch 2) → re-arm → fly → ctrl-c teardown (reverse order, zero leftover processes, pid marker removed).

luofang34:
All four review blockers addressed in 05179b5, rebased onto main (which now carries #121, resolving the `PILOTAGE_AVIATE_PROFILE` dependency):

1. **Post-spawn failure ownership** — pid-marker write failure now tears down the running stack; a spawned-but-never-ready FC replacement is killed before the error propagates (it is not yet in the supervised set, so the caller's teardown would miss it).
2. **Unconditional group KILL** — `terminate_group` no longer returns early when the leader exits during TERM grace; descendants that ignore TERM now always receive the group KILL.
3. **Port 0 rejected** on both `--port` and `--viewer-port`: an ephemeral bind that readiness checks and the printed URL never learn about cannot be advertised.
4. Restart counter uses `wrapping_add`.

Guardrail tests: TERM-ignoring grandchild dies after its leader exits; readiness failure tears down every started stage; failed FC restart kills the unready replacement; port 0 refused. Live rerun: reset mid-session (FC restart 1/3, shm epoch 2 + MAVLink clock epoch 2 reattach), ctrl-c teardown, **zero surviving processes**, pid marker removed.

luofang34:
Round-2 blockers addressed in 4706423:

1. **Ctrl-c during startup can no longer orphan the stack** — the SIGINT handler registers synchronously before the first child spawns; both the initial readiness waits and the FC-restart readiness wait race the cancellation source; on cancel, the not-yet-recorded child and every started stage are torn down. Startup cancellation is the typed `Cancelled` stop → clean exit.
2. **The proven host port is carried** — `ReadySignal::HostListening { port, certificate }`; the session fails closed on `PortMismatch` instead of printing a URL the host never bound; the viewer URL uses the proven port.
3. **Direct marker-write-failure guardrail** (`claim_supervisor`: teardown + drained children on write failure).
4. **Sleep-and-retry tests replaced with process-event synchronization** — each stage holds a fifo write end: open = started, EOF = the entire group is gone. New injected-cancellation tests (startup + restart) assert zero surviving groups; also covered: readiness failure, failed restart, TERM-ignoring grandchild, port mismatch.
5. CLI failures route through `tracing::error!`.

One live check outstanding: the ctrl-c-per-startup-phase matrix needs the machine's current manual session to close first (the stale-session guard correctly refuses a second stack); the fifo-EOF tests prove the same group-death outcome at the process level.

PR #129, opened 2026-07-17
MEDIA-01: a wedged stream write costs one frame, not the source (#126)

### LINKED ISSUE #126: Video source dies permanently when a client stalls one stream write
Observed live (session 07:49–07:52Z Jul 17): chase camera (source 1) logged 2,119 consecutive `video frame dropped: client writer still busy with the previous frame` over the final 70 s while source 0 kept streaming and the client stayed connected. A world/FC reset does not recover it — the wedged writer lives on the WebTransport connection, which reset never touches; only a page reload (reconnect) recovers.

Mechanism (`hosts/session-host/src/runtime/media.rs`): the per-(client, source) writer awaits `open_uni()` / `write_all()` / `finish()` with no deadline. A client that stops consuming one stream parks the write forever; the capacity-1 drop-to-latest handoff assumes the in-flight write eventually completes, so a permanently parked write turns drop-to-latest into drop-everything-forever for that source.

Fix: a per-frame write deadline — an expired write abandons its stream (reset toward the peer) and the writer proceeds with the next frame, so a wedged stream costs one frame, not the source. Stalls counted and logged.

luofang34:
## Problem
The per-(client, source) video writer awaited `open_uni`/`write_all`/`finish` with no deadline. A client that stops consuming one stream parks that write forever; the capacity-1 drop-to-latest handoff assumes the in-flight write eventually completes, so a parked write turned drop-to-latest into **drop-everything-forever** for that source. Observed live (session 07:49–07:52Z): 2,119 consecutive drops of the chase camera over 70 s while source 0 streamed on and the client stayed connected — and it survived a world/FC reset, because the wedged writer lives on the WebTransport connection, which reset never touches. Only page reload recovered.

## Fix
Each frame now writes under `FRAME_WRITE_DEADLINE` (2 s); an expired write is abandoned (dropping the in-flight future drops its stream, resetting it toward the peer) and the writer proceeds with the next frame. Stalls are counted and logged; an outright write error still ends the writer, since that connection is going away. The writer moved to `media/frame_writer.rs` (media.rs was over the 500-line limit) with the policy behind a write seam.

## Guardrails
Virtual-time tests on the seam: a hanging write must not block the next frame (`stalled_write_is_abandoned_and_the_next_frame_proceeds`); a write error ends the drain (`write_error_ends_the_drain`).

## Verified live
Rebuilt host, full session: both cameras painting through connect/arm/climb/land/disarm.

Closes #126

---
- <kbd>&nbsp;2&nbsp;</kbd> #129 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #121

### DIFF
--- a/hosts/session-host/src/runtime/media.rs
+++ b/hosts/session-host/src/runtime/media.rs
@@ -21,10 +21,12 @@ use tokio::sync::mpsc;
 use tokio::sync::mpsc::error::TrySendError;
 use tokio::task::JoinSet;
 use tokio::time::Instant;
-use tracing::{debug, error, warn};
+use tracing::{debug, warn};
 use wtransport::Connection;
 
-use crate::runtime::stream_tag::{FOURCC_MJPEG, VIDEO_FRAME_V2, frame_video_payload_v2};
+use frame_writer::client_writer;
+
+mod frame_writer;
 
 /// JPEG quality (1-100) for encoded camera frames. 75 balances size against
 /// visible quality for a teleop preview (ADR-0005: owned, tunable pipeline).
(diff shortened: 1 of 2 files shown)


luofang34:
Stacked on #121: the frame-writer test fixture builds LINK-04 measurement stamps (`role`/`integrity`), which only exist on the `link-04-source-roles` branch — the defect itself was observed and fixed against that build. Merging #121 first retargets this PR to main automatically.

luofang34:
Explicit reset fix (commit e932ff9).

You're right — `timeout()` dropped the write future, dropping the wtransport `SendStream`, and on 0.7.1/Quinn that's a graceful FIN attempt, not RESET_STREAM. A stalled peer never drains the FIN, so truncated streams accumulate against its stream allowance.

**Fix:** the stream is now opened outside the timed region and retained, so a deadline timeout calls `SendStream::reset(VarInt)` explicitly — RESET_STREAM, which drops buffered data and frees the peer's slot immediately — before the writer proceeds to the next frame. Confirmed against the wtransport 0.7.1 source: `reset(&mut self, VarInt)` is the RESET_STREAM path ("Locally buffered data is dropped, and previously transmitted data will no longer be retransmitted").

**Test:** open and write now sit behind `FrameChannel`/`FrameStream` seams, so the policy is tested with a mock stream that records calls — not the generic pending-future stub. `a_stalled_write_resets_once_and_the_next_frame_proceeds` proves a stalled write resets its stream **exactly once** and the next frame opens its own stream and finishes cleanly; I verified it fails if the explicit reset is removed. `a_write_error_ends_the_writer` proves an outright failure stops the writer (next frame never opened) and is not a reset.

**Warning:** now carries `client` and `source_id`.

Base is already main, so no stacking concern here.

luofang34:
Open is now inside the deadline (commit f7884d4), rebased on latest main.

You're right — the deadline bounded only the write, so a blocked `open_uni()` could still park forever and wedge the source, which is squarely the #126 failure (it includes stream opening). The per-frame budget now covers **both** phases under one absolute deadline via `timeout_at(deadline, …)`:
- **Open times out** → counted and skipped, writer continues to the next frame (no stream exists to reset).
- **Write times out after opening** → the retained stream is explicitly reset (RESET_STREAM), as before.

The stall warning carries a `stream_reset` field distinguishing the two.

**New virtual-time test** `a_stalled_open_is_skipped_and_the_next_frame_proceeds`: the first `open()` never completes; the frame is skipped with no reset, and the next frame opens its own stream and finishes cleanly. The existing stalled-write (resets exactly once) and write-error (ends the writer) tests are unchanged in intent, now driven through the open-behavior script.

Rebased on main and pushed; CI rerunning.

PR #131, opened 2026-07-17
LINK-05: judge a numeric by the status regime that governed it (#130)

### LINKED ISSUE #130: Panels flash valid/invalid: exact-instant status pairing strips interleaved lanes
With the estimate stream at 50/30 Hz and the status pair riding every emission, the host merges lanes into one sample per 10 ms tick — so a numeric group lawfully arrives alongside a status acquired a few ms later. The ingress demanded the numeric's acquisition equal the current status stamp's instant exactly; on every interleaved arrival it stripped a validly authorized lane (measured live: 90/300 samples left ingest with flags 12 or 3 while the wire carried 15), and the wasm honestly rendered each stripped frame as red-X — panels flashed correct/invalid at the ~10 Hz interleave beat, even at hover.

Fix: authorization regimes. Each accepted status opens a regime and the closed one is retained; a numeric is judged by the status that governed its acquisition instant (current, or the previous within its reign), bounded by the coherence budget against the current status. Identity/clock must match in every case; beyond the budget or across a reset it still fails closed — and a numeric acquired during a declared-unusable regime still gets nothing. The guard lives in the shared AvionicsIngress with the budget injected per profile, so simulation and physical clients get the same protection.

luofang34:
## Problem
With the estimate stream at 50/30 Hz and the status pair riding every emission (Aviate loopback rates), the host merges lanes into one sample per 10 ms tick — a numeric group lawfully arrives alongside a status acquired a few ms later. The ingress demanded exact acquisition-instant equality with the current status stamp; every interleaved arrival stripped a validly authorized lane. Measured live: **90/300 samples left ingest with flags 12 or 3 while the wire carried 15**; the wasm honestly rendered each stripped frame as red-X — panels flashed correct/invalid at the ~10 Hz interleave beat, even at hover.

## Fix
Authorization regimes: each accepted status opens a regime and retains the one it closes; a numeric is judged by the regime that governed its acquisition instant (current, or the previous within its reign), bounded by the coherence budget against the current status. Identity/clock must match everywhere; beyond the budget or across a source reset it fails closed; a numeric acquired during a declared-unusable regime still gets nothing (the round-2 contract that a later good status cannot retroactively bless it is preserved and pinned by tests).

The guard lives in the shared `AvionicsIngress` with the budget injected per profile (`maximumSkewNanos`) — simulation and physical clients enforce the same rule.

## Verified live
Shimmed the new logic into a running session: **400/400 post-ingest samples hold flags 15** (previously 30% stripped); the panel canvas region that flipped between exactly two states (correct/red-X, 42 toggles per 4 s) shows continuous scene motion.

Closes #130

### DIFF
--- a/clients/web/main.js
+++ b/clients/web/main.js
@@ -55,9 +55,10 @@ import { CalibrationRegistry, loadCalibrationRegistry } from "./calibration.js";
 
 const VEHICLE_ID = 1n; // demo fixture: the single Gazebo vehicle this host serves.
 const INSTRUMENT_SOURCE_ID = 1n; // explicit simulator adapter source; never first-packet selection.
-// Aviate publishes attitude at 10 Hz and kinematics at 4 Hz. This simulator
-// profile admits one kinematics period plus transport jitter; an aircraft
-// profile must derive its own limit from the intended function.
+// Aviate publishes attitude at 50 Hz and kinematics at 30 Hz. This simulator
+// profile admits several stream periods plus transport jitter; an aircraft
+// profile must derive its own limit from the intended function. The same
+// budget bounds status-to-numeric authorization pairing in the ingress.
 const SIM_COHERENCE_LIMIT_NS = 300_000_000n;
 const MOTION_SCOPE = "vehicle.motion";
 const CONTROL_HZ = 30; // continuous control send rate; superseded samples are droppable (ADR-0011).
--- a/clients/web/telemetry-ingress.js
+++ b/clients/web/telemetry-ingress.js
@@ -329,12 +329,28 @@ export class AvionicsIngress {
   }
 
   applyStatusDowngrade(avionics) {
+    const incomingValidFlags = avionics.validFlags >>> 0;
+    const incomingQuality = avionics.quality >>> 0;
+    // A duplicate status stamp may only tighten authorization, never
+    // restore it — so the downgrade must fold into the regime itself,
+    // not only the published flags. Otherwise a later numeric bearing
+    // this same status stamp would consult the pre-downgrade regime and
+    // reverse a fail-closed decision. The fold is monotone (bitwise-and
+    // of flags, max of quality), so re-applying the status that opened
+    // the regime is a no-op and successive duplicates only ever tighten.
+    if (this.statusRegime !== null) {
+      this.statusRegime = Object.freeze({
+        stamp: this.statusRegime.stamp,
+        validFlags: (this.statusRegime.validFlags & incomingValidFlags) >>> 0,
+        quality: Math.max(this.statusRegime.quality, incomingQuality),
+      });
+    }
     if (!this.hasEstablishedAuthorization()) {
       this.failClosedAuthorization();
       return;
     }
-    this.validFlags = (this.validFlags & (avionics.validFlags >>> 0)) >>> 0;
-    this.quality = Math.max(this.quality, avionics.quality >>> 0);
+    this.validFlags = (this.validFlags & incomingValidFlags) >>> 0;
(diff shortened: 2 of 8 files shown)


luofang34:
Reversal hole fixed (commit d735a5b), plus evidence trace (commit 2d5f6e7).

**The bug:** `applyStatusDowngrade` tightened only the published flags, not `statusRegime`. A duplicate status carrying a downgrade left the regime at its stale-good value, so a later numeric bearing that same status stamp consulted the pre-downgrade regime and reversed the fail-closed decision. Reproduced exactly as described: good status + attitude → flags 3; duplicate unusable status → flags 0, quality 2; a **newer** attitude (advanced acquisition, still within the skew budget) with the good payload → flags back to 3.

**Fix:** the downgrade now folds into the regime itself — `validFlags &= incoming`, `quality = max(existing, incoming)` — whenever a duplicate status stamp publishes it. Monotone, so re-applying the status that opened the regime is a no-op and successive duplicates only tighten. The later numeric consults the tightened regime and stays unauthorized.

**Test:** `testDuplicateStatusDowngradeIsNotReversibleByANewerNumeric` drives a fresh numeric sequence (acquired 50n later — genuinely new, not an age-frozen duplicate) after the downgrade. Verified it **fails without the fix** (`3 !== 0`) and passes with it. My first attempt used the same acquisition instant and the ingress silently dropped it as a non-advancing sample — that's why the exact-instant repro was needed.

**Evidence (third ask):** LINK-AUTHZ-005 is now a traced requirement in `docs/link/evidence-graph.evg` — requirement → design → impl → verification case → recorded result, wired to the review slot. The result is recorded against CFG-LINK-AUTHZ (commit d735a5b) and its case binds the JS regression test. To make that resolve honestly rather than as a broken node, the evidence selector resolver now resolves JavaScript test selectors (definition + runner registration, comment/string-decoy-safe, unit-tested) — the browser suites are real CI gates. The baseline guardrail verifies the test blob belongs to d735a5b (confirmed it fails on a tampered digest).

**Merge note:** like #121, this references a lane commit (d735a5b) from evidence — please **merge-commit, not squash**, so baseline resolution stays valid.

luofang34:
Both blockers fixed.

**1. Previous-regime reversal (commit 337e9b8).** You're right — selecting the previous regime alone ignored a since-downgraded current status. The current status regime is now a **monotonic authorization ceiling**: its validFlags intersect and its quality raises every lane, whichever regime governed the acquisition instant. A good current regime caps nothing (the legitimate interleave case is untouched); a downgraded one caps everything. Your exact repro is the regression test `testDowngradeIsNotReversibleThroughThePreviousRegime` — prev good@1000, curr good@1200, duplicate curr downgrades to (0,2), fresh numeric acquired@1100 selects the previous regime but is ceilinged to flags 0. Verified it returns flags 3 without the ceiling, 0 with it (and confirmed the numeric is genuinely *admitted*, so the ceiling — not rejection — is what caps it).

**2. JS selector fail-open (same commit).** `defines_js` is rewritten: a test resolves only with a **top-level** `function <symbol>(` definition AND membership in a **top-level** `for (… of [ … ])` runner array. A proper lexer blanks every comment and string literal (single-, double-, and backtick-quoted incl. `${…}`) first. Negative tests cover every case you listed: self-reference without registration, reference from another helper (a call, and a helper's own for-of array), nested function definitions, and single-quoted + template-literal decoys — plus a substring-of-registered-name guard.

**Evidence re-recorded (commit 4436773)** against the new baseline 337e9b8 (the test file blob changed), trace resolves both ways, review stays honestly pending.

Merge-commit (not squash) still required — evidence references lane commit 337e9b8.

PR #137, opened 2026-07-18
Viewer: auto-recover a dropped session and hold neutral control off-focus

luofang34:
## What broke

Reported as *"xbox controller stopped working and the sim lost responsiveness after a while."* Two coupled demo-viewer control defects, one behind the other:

1. **The session silently died and never came back.** When the browser backgrounds a tab it freezes the control loop, so its keepalive datagrams stop and the WebTransport session idle-drops (a host-initiated or errored close does the same). `handleTransportClosed` only retired the presentation — control, gamepad, and video went dead with no path back short of the user noticing and clicking **Connect**.
2. **A frozen stick kept flying.** Losing window focus *freezes* the Gamepad API (Chrome stops updating axis values) and stops keyboard events, but the control loop kept sending the **last** stick position — a runaway command with no live input behind it.

## The fix

**Auto-recovery.** `handleTransportClosed` and the `connect()` failure path now schedule a bounded, visibility-gated reconnect:
- exponential backoff, base 1 s, capped at 15 s, up to 6 attempts;
- only while the user still wants to be connected (they clicked Connect and never disconnected), no session is already active, and the tab is visible — a hidden/frozen tab would just drop again, so recovery waits and fires on `visibilitychange` when the user returns;
- a successful handshake and a manual **Connect** both reset the backoff.

**Neutral-on-blur.** While `document.hasFocus()` is false the control loop holds neutral axes (rover: throttle/yaw = 0; flight: roll/pitch/throttle/yaw = 0), suppresses arm/reset/fpv edges, and the readout says *"control paused — focus the window to fly (holding neutral)."*

## Guardrail

The reconnect **decision** logic — the backoff curve and the attempt/give-up gating — is extracted to a pure, side-effect-free `clients/web/session-recovery.js` and covered by `session-recovery.test.mjs` (wired into CI), so the policy is verified apart from the WebTransport plumbing in `main.js`.

## Verification

- `node clients/web/session-recovery.test.mjs` — 11/11 pass.
- Full non-browser web suite (15 suites) green; `node --check clients/web/main.js` clean.

The `#103` `createWritable` cross-engine hunk stays uncommitted, unchanged by this branch.

> Stacked on #136 (video log-spam dedup + uni-stream reader recovery); both touch `clients/web/main.js` and the viewer CI steps.

---
- <kbd>&nbsp;2&nbsp;</kbd> #137 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #136

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,9 @@ jobs:
       - name: viewer video-diagnostics decision logic (log dedup + reader restart)
         run: node clients/web/video-diagnostics.test.mjs
 
+      - name: viewer connection auto-recovery decision logic (backoff + reconnect gating)
+        run: node clients/web/session-recovery.test.mjs
+
       - name: wasm wire-decode conformance (REN)
         # The wasm exports the viewer decodes with must agree with the JS
         # reference decoders on layout, numeric kinds, and the capture-identity
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,9 @@ jobs:
       - name: browser video capture-identity tests (HUD-01)
         run: node clients/web/video-identity.test.mjs
 
+      - name: viewer video-diagnostics decision logic (log dedup + reader restart)
+        run: node clients/web/video-diagnostics.test.mjs
+
       - name: wasm wire-decode conformance (REN)
         # The wasm exports the viewer decodes with must agree with the JS
         # reference decoders on layout, numeric kinds, and the capture-identity
--- a/clients/web/main.js
+++ b/clients/web/main.js
@@ -57,6 +57,7 @@ import {
   shouldLogReadFailure,
   restartVerdict,
 } from "./video-diagnostics.js";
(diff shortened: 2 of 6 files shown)


luofang34:
Superseded by the review-requested split. #137 combined session reconnection with input-loss safety and had four safety blockers (auto-resuming motion authority, backoff reset before bootstrap, a held button edging across reconnect, and a 30 Hz neutral stream that kept command freshness alive), plus the missing host-side guardrail.

Replaced by three focused PRs:
- **#142** — Viewer: auto-reconnect the transport without resuming control (reconnection: control stays suspended, backoff-after-bootstrap, jitter + slow capped retry, non-retryable classification; fake-time/fake-WebTransport lifecycle tests).
- **#143** — Viewer: relinquish control authority on focus loss; no held-button edge (input-loss safety: stop publishing → host expires the holder, held-button-no-edge, blur clears edge state).
- **#141** — Host: expire silent holders and neutralize the vehicle on link loss (the holder-liveness watchdog + live `set_link_loss_policy` enactment — the deeper guardrail).

Closing in favor of those.

PR #144, opened 2026-07-18
Flight: brake to a stop before capturing the stick-release hold point

### LINKED ISSUE #123: Yoke release overshoots back: brake-then-hold captures the hold point at release, before braking
## Symptom (SITL, identified while flying manually)

Release the yoke after a translation and the vehicle does not simply stop: it keeps moving, brakes, then visibly flies **back** toward where the sticks were released (~1–2 m at full stick speed) before settling. Same effect vertically after releasing the throttle stick, at smaller amplitude.

## Root cause — hold point captured at release, with no brake phase

`adapters/aviate/src/uplink.rs`, `send_stick_frame`:

```rust
// DJI brake-then-hold: with every stick centered, capture the
// current position once and stream a position setpoint. ...
if !sticks_active {
    let hold = *self.hold_pos_ned.get_or_insert(current_pos_ned_m);
    ...
    encode_position_setpoint(..., hold, ...)
```

The comment promises **brake-then-hold**, but the code implements **hold-at-release**: `hold_pos_ned` is captured on the *first centered-stick frame*, while the vehicle is still moving at up to `MAX_HORIZONTAL_MPS = 3.0 m/s`. From that instant the FC is commanded `PositionHold` on a point the vehicle is about to overrun. The rest is kinematics, executed faithfully by the FC:

1. At capture, position error ≈ 0, so the FC's sqrt position controller commands vel_sp ≈ 0 and the velocity loop brakes at its tilt-cap authority (`vel_max_roll_pitch = 0.35 rad` → ≈ g·tan(0.35) ≈ **3.5 m/s²**).
2. Braking from 3 m/s covers v²/2a ≈ **1.3 m** past the hold point — plus tilt-reversal lag (the vehicle is tilted *into* the motion at release and must swing through level to the opposite tilt), in practice ~1.5–2 m of overtravel.
3. The position loop now sees 1.5–2 m of error pointing **back** at the capture point (linear region: `pos_p = 0.5` → ~0.75–1.0 m/s return command) and flies the vehicle back. That return leg is the reported "overshoots back".

Two smaller contributors, same mechanism:

* The captured `current_pos_ned_m` comes from the telemetry estimate (`current_pose()`), so the hold point additionally lags the true release position by the downlink latency — the pull-back target is slightly *behind* even at capture time.
* Vertical: throttle release captures altitude while climbing/descending at up to 1.5 m/s → ~0.4 m of altitude bounce-back.

## What it is NOT (checked in aviate-core)

* Not the FC position loop: the sqrt controller (`control/position.rs`) guarantees no overshoot around **its** setpoint approached within the braking envelope — it cannot un-happen momentum toward a point commanded behind the motion.
* Not integrator windup: the velocity loop's conditional anti-windup freezes accumulation while saturated (`control/velocity.rs`), and `vel_i = 0.05` is small.
* Not the accel feedforward spike at the velocity→position mode switch: `vel_accel_ff = 0.0` in `x500_defaults` (disabled).

## Fix sketch (ground side only, no control gains — consistent with the documented split)

Implement the brake phase the comment already claims:

1. While sticks are centered **and** `|vel_ned|` is above a threshold (telemetry already carries `vel_ned_mps`), stream **zero-velocity setpoints** — exactly what `send_neutral` does; the FC's velocity mode brakes to a hover on zero demand.
2. Only when `|vel_ned|` falls below the threshold (e.g. 0.3 m/s), capture `hold_pos_ned` and switch to streaming the position setpoint.
3. Keep `hold_pos_ned = None` reset on any stick deflection, as today.

This is the DJI/PX4 behavior ("smart hold": brake first, capture the hold point after the vehicle has stopped). Threshold-on-a-measurement is not a control gain; the hold loop itself stays on the FC.

## Acceptance

* SITL: full-stick translation at 3 m/s, release → the vehicle's position approaches its final hover point monotonically after the brake (no return leg beyond estimator noise; today it returns ~1.5–2 m).
* Same check on the vertical axis from a 1.5 m/s climb.
* Link-loss neutralize behavior (`send_neutral`) unchanged.

luofang34:
Fixes #123.

## Symptom / root cause

Release the yoke after a translation and the vehicle brakes, then flies **back** ~1.5–2 m: `send_stick_frame` captured the hold point on the first centered-stick frame while still moving at up to 3 m/s, so the FC was commanded PositionHold on a point the momentum overruns. (Full kinematic walkthrough in #123.)

## Fix — ground side only, no control gains

Centered sticks stream **zero-velocity setpoints** (the `send_neutral` frame), and the hold is captured only on **positive evidence of stillness**:

- Velocity is **independently validated data**: `current_pose()` returns it as `Option<[f32;3]>`, present only when the FC declared the velocity group valid (bit 3 — the same gate the planar speed projection uses) and every component is finite.
- Capture requires validated, finite velocity **at or below** 0.3 m/s. Missing, undeclared, or non-finite velocity keeps braking — "stopped" is never inferred from absent data (`NaN > 0.3` is false, which would otherwise capture the hold at full speed — the exact defect this PR prevents).
- The captured hold is sticky against velocity blips; any stick deflection returns to velocity mode. `send_neutral` (link-loss neutralize) untouched.

The threshold is on a measurement, not a control gain; the hold loop stays on the FC (LINK-04 gating: OperationalEstimate only, simulation truth ineligible).

## Deterministic timing

`FlightUplink` runs on an injectable clock (system in production, manually advanced in tests), so the post-arm quiet window and slew-limiter dt are exercised **without real-time sleeps** — the whole suite runs in ~10 ms.

## Guardrails (fake FC socket, manual clock)

- brake at |vel| ≈ 5 m/s (mask 2503, zero demand), purely vertical 1 m/s residual still brakes, capture below threshold (mask 2552 at the fixture pose), hold survives a velocity blip;
- **position-valid / velocity-invalid** (bit 3 cleared, even at numerically zero velocity) never captures — braking continues until validity plus stillness;
- **NaN, +∞, −∞** components each keep braking;
- **threshold boundary**: 0.301 m/s brakes, 0.299 m/s captures.

## Status / why draft

- [x] `cargo test -p pilotage-adapter-aviate` — 89 passed (no sleeps)
- [x] fmt, clippy `--all-targets -- -D warnings`, `check-structure.sh` clean; verified in an isolated worktree
- [x] SITL trajectory acceptance per #123 — **recorded** (see the results comment): vertical bounce-back 0.041 m (baseline ~0.4 m); horizontal worst backward excursion 0.49 m of damped hold-loop ringing with no directed return leg (baseline 1.5–2 m)

### DIFF
--- a/adapters/aviate/src/adapter.rs
+++ b/adapters/aviate/src/adapter.rs
@@ -7,8 +7,8 @@ use std::time::Duration;
 
 use pilotage_adapter_api::{
     AdapterCapabilities, ApplyOutcome, Disposition, ExecutionMode, LinkLossPolicy, RejectReason,
-    ScopeDescriptor, SourceIncarnation, SourceRole, StepBudget, StepOutcome, TelemetryBatch,
-    TelemetrySample, VehicleAdapter, VehicleDescriptor, VideoSource,
+    ScopeDescriptor, SourceIncarnation, StepBudget, StepOutcome, TelemetryBatch, TelemetrySample,
+    VehicleAdapter, VehicleDescriptor, VideoSource,
 };
 use pilotage_protocol::{
     ButtonEdge, LogicalAxisId, LogicalButtonId, ScopeId, ScopedControlFrame, VehicleId,
@@ -31,9 +31,7 @@ mod shm_sampling;
 mod sources;
 mod startup;
 use control::{flight_button_pressed, normalized_flight_sticks, rejected_control};
-use sampling::{
-    mavlink_batch, measurement_pair_is_coherent, measurement_pair_supports_pose, yaw_of,
-};
+use sampling::mavlink_batch;
 use shm_sampling::ShmSource;
 use sources::{ArmReport, EstimateSource, fc_state_sample};
 
@@ -182,29 +180,10 @@ impl AviateAdapter {
         self
     }
 
-    /// The vehicle's current measured yaw (radians clockwise from
-    /// north) and NED position, FROM THE FC OPERATIONAL ESTIMATE ONLY
-    /// (LINK-04): simulation truth is never eligible to seed command
-    /// construction, so without a live authorized estimate there is no
-    /// pose and state-dependent control is rejected instead of
-    /// borrowing truth.
-    fn current_pose(&mut self) -> Option<(f32, [f32; 3])> {
-        let latest = self.estimate.as_ref()?.state.lock().ok()?;
-        latest.estimator_status_stamp()?;
-        let attitude = latest
-            .attitude
-            .filter(|update| update.received_at.elapsed() <= WITHHOLD_AFTER)
-            .filter(|update| update.stamp.role == SourceRole::OperationalEstimate)?;
-        let kinematics = latest
-            .kinematics
-            .filter(|update| update.received_at.elapsed() <= WITHHOLD_AFTER)
-            .filter(|update| update.stamp.role == SourceRole::OperationalEstimate)?;
-        if !measurement_pair_is_coherent(attitude, kinematics, latest.maximum_inter_group_skew_ms)
-            || !measurement_pair_supports_pose(attitude, kinematics)
-        {
-            return None;
-        }
-        Some((yaw_of(attitude.quat_wxyz) as f32, kinematics.pos_ned_m))
+    /// The bound uplink, for tests that drive its manual clock.
+    #[cfg(test)]
+    pub(crate) fn uplink_mut(&mut self) -> Option<&mut FlightUplink> {
+        self.uplink.as_mut()
     }
 
     fn validate_flight_frame(&self, frame: &ScopedControlFrame) -> Result<(), RejectReason> {
@@ -293,7 +272,7 @@ impl VehicleAdapter for AviateAdapter {
                 disposition: Disposition::Accepted,
             };
         }
-        let Some((current_yaw, current_pos)) = self.current_pose() else {
+        let Some((current_yaw, current_pos, current_vel)) = self.current_pose() else {
             return rejected_control(tick, RejectReason::MeasurementUnavailable);
         };
         let Some(uplink) = self.uplink.as_mut() else {
@@ -328,6 +307,7 @@ impl VehicleAdapter for AviateAdapter {
                 sticks[usize::from(YAW_AXIS)],
                 current_yaw,
                 current_pos,
+                current_vel,
             );
         }
         ApplyOutcome {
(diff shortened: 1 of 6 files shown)


luofang34:
## SITL trajectory acceptance — recorded results (per #123)

**Setup.** Full stack at this PR's head `0a4274f` via `cargo xtask sim` (worktree checkout; Gazebo Harmonic headless + Aviate SITL FC `sitl-gazebo-x500` + session-host release build, simulation profile, sim RTF ≈ 1.0). Flown through the production path: browser viewer keyboard control (`quad-pilot`, 30 Hz control loop → adapter → `FlightUplink` → FC). Ground truth recorded from the sanctioned external tap (`/aviate_gz_bridge_v3` seqlock reader via Aviate's `ConsumerSession`), one sample per physics step; velocities derived from position over sim time (median-smoothed). Raw trajectory: `pilotage-144-sitl-trajectory-20260718T230809Z.csv` (185,878 samples).

**Vertical (release from a full 1.50 m/s climb).**
- Measured climb rate: 1.50 m/s (median).
- Release at 40.86 m → apex 40.94 m: **+0.088 m braking overshoot** (≈ v²/2a, pure physics).
- Settle 40.88 m; **bounce-back below settle: 0.041 m** — vs the ~0.4 m defect baseline in #123. **PASS.**

**Horizontal (release from full-stick cruise), two legs, opposite directions.**
| leg | peak speed | brake distance past release | max backward excursion after stop | settle vs stop apex | crossings of final |
|---|---|---|---|---|---|
| 1 | 2.82 m/s | 1.24 m | −0.489 m | −0.243 m | 7 |
| 2 | 2.93 m/s | 1.69 m | −0.278 m | −0.148 m | 12 |

The defect signature — a single **directed return of 1.5–2 m** toward the point where the sticks were released (the hold captured at release) — is **absent**: the vehicle brakes 1.2–1.7 m past release (its physical stopping envelope), captures the hold near the stop point, and exhibits a *damped oscillation* about it (7–12 sign-changes around the final position, worst single excursion 0.49 m, settling within 0.25 m of the stop apex). That residual ringing is the FC position-loop's damping about any captured hold — it would be present even for a hold captured at perfect rest — and belongs to aviate-core gain tuning, not the ground-side capture logic this PR fixes. **PASS** against the #123 criterion (no return leg toward the release point; baseline 1.5–2 m).

**Link-loss neutralize** unchanged (`send_neutral` untouched; covered by the adapter unit suite, 89 tests at this head).

Reproduce: `cargo xtask sim` from this branch (viewer wasm bundle must be built or copied — it is a gitignored artifact), fly climb-release and cruise-release via viewer keys, record the shm tap.

Issue #168, opened 2026-07-19
GIM-04: PX4 Gimbal Protocol v2 backend + gz_x500_gimbal SITL rig


luofang34:
Part of #162 (GIM-01). Depends on the `vehicle.gimbal` scope (GIM-03).

## Goal

First gimbal backend: the PX4 adapter maps `vehicle.gimbal` frames to MAVLink Gimbal Protocol v2, tested end-to-end against PX4 SITL `gz_x500_gimbal`.

## Adapter mapping

- Declare `vehicle.gimbal` in `AdapterCapabilities` for PX4 vehicles.
- Claim gimbal primary control once (`MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE`) before the first setpoint; surface a typed rejection if the claim is denied.
- Rate frames → GMv2 rate setpoints (normalized demand × configured max rates → rad/s); reset edge → neutral (`MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW` NEUTRAL flag or zero-attitude setpoint — pick one, document why).
- Stale-demand safety: stop streaming rate setpoints when frames stop (PX4 zeroes a stale rate after 2 s; we must not rely on that alone).
- Ingest `GIMBAL_DEVICE_ATTITUDE_STATUS` as telemetry with a proper `SourceRole` stamp (available to clients; display is out of scope).

## SITL rig

- xtask px4-gz backend gains model selection so the equivalent of `make px4_sitl gz_x500_gimbal` runs under `cargo xtask sim`.
- E2E acceptance: launch sim, drive gimbal pitch/yaw via the LT+RStick path, observe gimbal joint motion in gz and attitude status on the wire; RStick-click returns it to neutral.

## Guardrails (same PR)

- Golden wire vectors for every new MAVLink message encode/decode in `pilotage-mavlink`.
- Unit tests for the claim→setpoint→neutral sequence and for setpoint-stream cutoff on frame loss.

## Out of scope

- Vehicle-yaw decomposition for pan-limited gimbals (x500_gimbal has a yaw axis; the LOS decomposer arrives with the `pilotage-gimbal` core).
- Turret/STorM32 backend (topology B), router (M5), tracker.

luofang34:
[...]

Issue #162, opened 2026-07-19
[TRACKING] GIM-01: camera payload pointing — LOS control, gimbal DOF, and track-assist


luofang34:
## Role

Tracking epic for camera payload pointing: gimbal degrees-of-freedom abstraction, line-of-sight (LOS) control, operator input ergonomics, and track-assist. Every implementation child is one independently revertible PR with its regression guardrail. Imaging control (zoom/snapshot/record) is explicitly out of scope here — it is a separate later family that ties into VID-03 (#102) encoder/capability negotiation.

### LOS setpoint, not camera angles

The operator- and tracker-facing model is a **desired line-of-sight in the world frame**, lock-mode by default: stick release holds LOS; the vehicle turning underneath is compensated automatically. Canonical units per SRC-01: rad, rad·s⁻¹.

A **DOF capability mask** declares what the physical/simulated gimbal can do. A shared **LOS decomposer** splits desired LOS into (vehicle yaw demand, gimbal pitch demand, optional gimbal roll for horizon hold) based on that mask:

- 1-axis (pitch) + vehicle yaw — first target; exercises the hard part (cross-scope coupling).
- 2-axis (+roll) — horizon hold on rolling airframes; decomposer sends less to the vehicle.
- 3-axis — only needed for fixed-wing or targets faster than vehicle yaw rate. No interface change at any step.

### Pointing is a control scope

`vehicle.gimbal` is a new host-published scope (`ScopeDescriptor`, ADR-0006/0008) with new logical axes (`gimbal_pitch`, `gimbal_yaw`) in the `pilotage-input` vocabulary. It rides `ScopedControlFrame` — lease, generation fencing, override — unchanged. No parallel arbitration mechanism.

### Cross-scope yaw coupling

A pan command on a pitch-only gimbal needs vehicle yaw:

- **Same-holder path (now):** the operator pointing the camera already holds `vehicle.motion`; in gimbal mode their yaw channel carries the decomposed LOS yaw. Input routing, not an authority question.
- **Tracker path (later):** an autonomous tracker is a control participant acquiring authority through `AuthorityEngine`. Pilot stick motion → override → track breaks, for free, per ADR-0006/0010.

### Same control for SITL and real

The Pilotage wire stays engine-agnostic; mapping lives in adapters (ADR-0008):

[...]

PR #169, opened 2026-07-19
GIM-04: PX4 Gimbal Protocol v2 backend + gz_x500_gimbal SITL rig

luofang34:
Refs #168, part of #162 (GIM-01).

First gimbal backend: the PX4 adapter advertises a second, independently
leased control scope — `vehicle.gimbal` (ADR-0006 vocabulary; supersedes the
`payload.gimbal` working name in the issue) — and maps it to MAVLink Gimbal
Protocol v2, tested live against PX4 SITL `gz_x500_gimbal`.

## What

- **pilotage-mavlink**: GIMBAL_MANAGER_SET_ATTITUDE (282) rate-demand encoder
  (all-NaN attitude quaternion, horizon + yaw-follow flags),
  GIMBAL_DEVICE_ATTITUDE_STATUS (285) decode with MAVLink-v2 zero-extension,
  CRC_EXTRA values verified against PX4's generated common-dialect headers.
  Outbound frames ride the receive link's own socket via a new mpsc lane: the
  FC's GCS instance retargets its stream to the last peer that spoke, so a
  second socket would steal the telemetry stream. Command acks and gimbal
  device status are cached on LinkState.
- **PX4 adapter**: `vehicle.gimbal` scope (pitch/yaw LOS rate demands +
  recenter edge). Primary control is claimed for the codec's GCS identity and
  re-asserted at 1 Hz while demands flow — PX4 silently drops non-primary
  SET_ATTITUDE — and a fresh CONFIGURE denial rejects pointing frames with a
  typed reason. A demand stream silent past 300 ms is cut off with one
  zero-rate setpoint (the FC's own stale fallback is 2 s). Pointing is
  deliberately outside the flight gate chain and keeps working where flight
  control is rejected.
- **xtask px4-gz + world**: boots `PX4_SYS_AUTOSTART=4019` /
  `gz_x500_gimbal`; the flight-deck world statically includes
  `model://x500_gimbal` under the same `x500_0` name (merge-include keeps
  `x500_0::base_link` valid for the camera-rig weld).

## Guardrails

- Golden wire-layout lockdown for the 282 encoder; 285 decode incl. truncated
  payloads; claim-before-stream, claim re-assert cadence, one-shot stale
  cutoff, neutral command, typed claim-denial rejection; xtask planning test
  pins the airframe env and a world test pins the model include (a plain-x500
  world boots cleanly and then ignores every gimbal demand silently).

## Verification

- Full local CI: fmt, clippy `-D warnings`, workspace tests, doc gates,
  release build.
- Live SITL E2E against `gz_x500_gimbal` (standalone gz + PX4, probe using
  these exact codec functions): CONFIGURE accepted, pitch leg 0 to -1.271 rad
  on a -0.5 rad/s stream, yaw leg +1.49 rad, neutral recentered to
  level/vehicle-forward; acks `[(1001,0)x5, (1000,0)]`.

### DIFF
--- a/adapters/aviate/src/adapter.rs
+++ b/adapters/aviate/src/adapter.rs
@@ -356,6 +356,7 @@ impl VehicleAdapter for AviateAdapter {
                     avionics: None,
                     sim_truth: truth,
                     fc_state,
+                    gimbal: None,
                 }],
             };
         }
--- a/adapters/aviate/src/adapter/sampling.rs
+++ b/adapters/aviate/src/adapter/sampling.rs
@@ -163,6 +163,7 @@ pub(crate) fn mavlink_batch(vehicle: VehicleId, state: &Arc<Mutex<LinkState>>) -
             avionics,
             sim_truth: None,
             fc_state: None,
+            gimbal: None,
         }],
     }
 }
--- a/adapters/aviate/src/uplink/fc_replies.rs
+++ b/adapters/aviate/src/uplink/fc_replies.rs
@@ -29,7 +29,9 @@ impl FlightUplink {
                 }
                 match *message {
                     pilotage_mavlink::codec::FcMessage::Heartbeat { armed: a } => armed = Some(a),
-                    pilotage_mavlink::codec::FcMessage::CommandAck { command, result } => {
+                    pilotage_mavlink::codec::FcMessage::CommandAck {
+                        command, result, ..
+                    } => {
                         if result == 0 {
                             info!(command, "FC accepted command");
                         } else {
--- a/adapters/gazebo/src/adapter.rs
+++ b/adapters/gazebo/src/adapter.rs
@@ -261,6 +261,7 @@ fn telemetry_from_odometry(vehicle: VehicleId, odom: &BridgeOdometry) -> Telemet
         avionics: None,
         sim_truth: None,
         fc_state: None,
+        gimbal: None,
     }
 }
 
--- a/adapters/px4/src/adapter.rs
+++ b/adapters/px4/src/adapter.rs
@@ -23,6 +23,8 @@ use crate::uplink::Px4Uplink;
 
 mod camera;
 mod control;
+#[cfg(test)]
+mod gimbal_tests;
 mod pointing;
 mod sampling;
 #[cfg(test)]
@@ -124,13 +126,22 @@ impl Px4Adapter {
     pub async fn start(vehicle: VehicleId, config: Px4Config) -> Result<Self, Px4AdapterError> {
         let link_config = config.link_config();
         let incarnation = pilotage_adapter_api::SourceIncarnation::new(rand_incarnation());
-        let link = MavlinkLink::start(link_config, incarnation).await?;
+        let mut link = MavlinkLink::start(link_config, incarnation).await?;
         let state = link.state();
-        let gimbal = Some(crate::gimbal::Px4GimbalControl::new(
-            link.outbound(),
-            link_config.system_id,
-            link_config.component_id,
-        ));
+        // The gimbal path is wired only when the vehicle is declared to
+        // carry one: a bare airframe advertises no `vehicle.gimbal`
+        // scope, so a client cannot lease a payload it cannot point. The
+        // rate lane is taken exactly once here, so it is always present
+        // on this first (and only) take.
+        let gimbal = match (config.gimbal, link.take_gimbal_rate_sender()) {
+            (true, Some(rates)) => Some(crate::gimbal::Px4GimbalControl::new(
+                link.command_sender(),
+                rates,
+                link_config.system_id,
+                link_config.component_id,
+            )),
+            _ => None,
+        };
         let uplink = match Px4Uplink::new(config.command_endpoint) {
             Ok(mut uplink) => {
                 uplink.set_expected_source(link_config.system_id, link_config.component_id);
@@ -353,6 +364,11 @@ impl VehicleAdapter for Px4Adapter {
     fn sample_telemetry(&mut self) -> TelemetryBatch {
         self.observe_arm_report();
         let fc_state = sampling::fc_state_sample(self.arm, self.arm_incarnation, self.started_at);
+        let gimbal_attitude = self
+            .gimbal
+            .is_some()
+            .then(|| self.gimbal_attitude())
+            .flatten();
         let mut batch = self
             .estimate
             .as_ref()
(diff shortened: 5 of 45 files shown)


luofang34:
Review repairs pushed (commit c409fa3), all five blocking findings addressed:

1. **Single-sequence outbound.** The link task now owns the socket's one MAVLink sequence and does all encoding. Two typed lanes replace the pre-encoded `mpsc<Vec<u8>>`: an ordered `OutboundCommand` queue (claims, recenters) and a latest-value `watch<GimbalRateDemand>` lane (rate setpoints coalesce under backpressure). `Px4GimbalControl` no longer holds a second counter, so interleaved sends can no longer produce duplicate/backward sequence numbers.\n2. **Truthful apply.** A claim/recenter refused by a full or closed lane now returns a typed `Rejected(Other)` instead of `Accepted`; rate demands drop-to-latest rather than accumulating stale.\n3. **Durable claim-denial.** `COMMAND_ACK` decodes and validates `target_system`/`target_component`; the gimbal CONFIGURE verdict lives in its own `gimbal_configure_ack` slot so a later `SET_MESSAGE_INTERVAL` (511) ack cannot bury a 1001 denial. Regression tests: `an_unrelated_ack_does_not_bury_a_gimbal_claim_denial` (511-after-1001) and `an_ack_addressed_to_another_endpoint_is_ignored`.\n4. **Explicit capability.** The gimbal path is wired only under `Px4Config::with_gimbal` (`PILOTAGE_PX4_GIMBAL`, set by the px4-gz backend). A bare airframe advertises no `vehicle.gimbal` scope — test `a_vehicle_without_a_gimbal_advertises_no_gimbal_scope`.\n5. **Client-visible telemetry.** `GimbalAttitude` joins `TelemetrySample` (proto + adapter-api, boxed to keep the routing enum within the large-variant budget) and `sample_telemetry` publishes the cached device attitude under the video-capture source role.\n\nGround-truthed live against gz_x500_gimbal after the refactor: claim from the ordered lane (PX4 primary 255/190), LT+stick tilts the HD camera, R3 recenters. Full local CI green (fmt, clippy -D warnings, workspace tests, doc, structure); verified self-contained in an isolated worktree.

luofang34:
Second-review repairs pushed (5bd4a3b + 1fe38bf). All five findings verified against the code and fixed:

1. **Claim-before-first-setpoint.** The outbound `select!` is now `biased;` with the command lane first, so a queued CONFIGURE always reaches the wire before the rate setpoint enqueued after it — an unbiased select could transmit the rate first.
2. **Truthful apply / honest logging.** A rate demand whose due claim could not be enqueued is no longer published (the sender lacks primary control), and the stale-stream cutoff only logs success when the zero-rate setpoint actually reached the lane.
3. **Honest device provenance.** New `SOURCE_ROLE_PAYLOAD_DEVICE` replaces the misleading `VideoCapture`; the stamp carries the device's own `time_boot_ms` under the vehicle-boot clock instead of host receive time. `GIMBAL_DEVICE` flags and `failure_flags` now travel on the sample/wire instead of being dropped.
4. **Gimbal-only telemetry survives.** `sample_telemetry` now carries a pose-less sample when no coherent avionics group is present, so gimbal-device (and FC-state) reports still reach clients. Regression tests: role/clock stamping, gimbal-only survival, failure-flag propagation.
5. **Evidence re-recorded, not hash-swapped.** The four stale LINK-04 results were re-baselined against the gimbal-telemetry commit: each suite re-run (all pass), run artifacts regenerated, source+output digests recomputed. `--require-resolvable` is green again; REVIEW-LINK04 stays PENDING as requested.

Full local CI green (fmt, clippy -D warnings, 68 test binaries, doc, structure, evidence gate); verified self-contained in an isolated worktree. Re-verified live against gz_x500_gimbal after the biased-arbiter refactor.

Renaming to the final `vehicle.gimbal` vocabulary across #162/#167/#168 (from the `payload.gimbal` working name) — will do that on the tracking issue before closing.

luofang34:
## Round-3 review — the four gaps are addressed

All four verified against the code before fixing; each fix lands with its guardrail.

**1. Gimbal telemetry now actually reaches the client** (`982de81`)
`decodeTelemetrySample` stopped at `fc_state` (field 8), so the host's field-9 gimbal lane was decoded by nobody. Added a role-gated `decodeGimbalAttitude` (`wire.js`) that surfaces orientation, angular rates, device flags, and failure flags, and — like the other stamped lanes — returns `null` unless the stamp carries the payload-device role, so a mislabeled lane can never drive the camera view or be read as FC state. `wire.test.mjs` covers the flags/failure-flags/stamp surface, the unstamped-is-null case, and the wrong-role gate.

**2. Payload-device reset/epoch discipline** (`83df3e9`)
The stamp borrowed the FC arm incarnation with a constant epoch and carried `time_boot_ms` as sequence + acquisition time. A gimbal/FC reboot regresses `time_boot_ms`, so time ran backwards under a fixed identity+epoch — indistinguishable from a stale replay. The stamp is now minted through a dedicated `GimbalStamp` discipline as the device report is ingested: a stable gimbal-source incarnation (kept apart from the arm incarnation, so a re-arm is not read as a reboot) plus an epoch that turns over whenever the boot time fails to advance. Within any one (identity, epoch) acquisition time is monotonic; a reboot is a visible epoch change. Regression test drives **5000 → 6000 (one epoch) → reboot 10 ms (next epoch) → re-sample (same epoch)**.

**3. Socket-level proof of CONFIGURE-before-rate** (`ccf58cb`)
The prior tests only inspected the command and rate lanes in isolation — never the arbiter that interleaves them. New `link/arbiter_tests.rs` binds a real UDP socket for the FC, queues a CONFIGURE claim **and** a rate demand so both are pending before the `biased` select first runs, and asserts the FC receives the `COMMAND_LONG` CONFIGURE (command 1001) before the `GIMBAL_MANAGER_SET_ATTITUDE` rate frame. An unbiased select would let the rate race ahead and PX4 would drop it as non-primary.

**4. Evidence — resolvable again; verdict still INVALID by design** (`2d21d48`)
The workspace rebase onto the latest `main` rewrote gim-04's SHAs, orphaning the recorded baseline `5bd4a3b` and reddening the hard `--require-resolvable` gate. Per the graph's MERGE/REBASE RULE the four telemetry/capture results are re-anchored to the durable reachable tip; the bound test files are byte-identical there and re-running each suite reproduces the recorded pass exactly (5/4/7/1), so only the provenance headers and output-digests moved.

To be explicit about what green means here: the evidence-trace-gate is only the **hard resolvability gate**. `REVIEW-LINK04` remains a **PENDING** review slot, so LINK-04's actual verdict stays **INVALID** until the independent review lands. **#168 and REVIEW-LINK04 stay open**; a green check here must not be read as "LINK-04 verified."

Issue #167, opened 2026-07-19
GIM-03: vehicle.gimbal scope, gimbal axes, and LT+RStick quasimode


luofang34:
Part of #162 (GIM-01).

## Goal

Add the `vehicle.gimbal` control scope end-to-end on the client side: scope vocabulary, gimbal logical axes, and the operator quasimode.

## Behavior (owner decision, 2026-07-19)

- **LT held + right stick** = gimbal rate control (stick deflection → LOS angular rate; release = hold). LT is the modifier **by owner decision**, superseding the LB recommendation in #162.
- While LT is held as the gimbal modifier, **LT's climb/descend contribution is gated to zero** (RT still climbs). This is a deliberate flight-control behavior change: the descend channel is unavailable while actively steering the gimbal. It must be visible in the control-edge/HUD state, not silent.
- **Right-stick click (R3)** = gimbal reset to neutral (one-shot button edge, not a level).
- Right stick continues to drive `vehicle.motion` when LT is not held; transitions in/out of the quasimode must neutralize the axes they leave behind (no stale yaw/pitch demand latching into either scope).

## Scope semantics

- `vehicle.gimbal` frames ride `ScopedControlFrame` unchanged: own generation fencing, sequence, profile revision.
- Axes within the scope: pitch + yaw rate demands in `[-1, 1]` normalized; canonical conversion to rad/s happens adapter-side (SRC-01).
- Reset is a button edge in the payload, not a magic axis value.

## Guardrails (same PR)

- Tests that quasimode entry/exit neutralizes both scopes' axes.
- Test that LT's throttle contribution is zero exactly while the quasimode is active.
- Wire/decode round-trip coverage for the new scope string and axes.

## Out of scope

- PX4/MAVLink mapping (GIM-04).
- Gimbal attitude display, tracker, geolocation.

luofang34:
[...]

Issue #109, opened 2026-07-15
CTRL-01: map logical controls to typed vehicle intents and make scheme changes fail-safe


luofang34:
## Problem

The current adapter advertises numeric axes under one broad `vehicle.motion` scope. The same roll/pitch/throttle/yaw IDs change meaning when a button toggles between velocity flight and attitude/direct-thrust flight. Arm, disarm, FPV toggle, and simulation reset are numeric button IDs interpreted inside the adapter.

This hides semantic and authority changes behind device positions, makes capability negotiation weak, and bakes quadcopter assumptions into a link intended to grow toward other aircraft and spacecraft.

## Ownership and layers

Pilotage owns:

1. logical controls produced by INPUT-01;
2. a versioned **ControlScheme** mapping them to typed intents/actions;
3. a vehicle binding that admits only intents advertised by the adapter/vehicle capability.

Aviate owns the typed command meaning after transport ingress and its verified-command boundary.

## Required change

- Define typed intent families with explicit units and reference frames, initially including velocity/position-hold and attitude/thrust or body-rate control.
- Define typed discrete actions such as arm/disarm/mode request; raw button numbers must not cross the adapter/vehicle-link boundary as command meaning.
- Advertise supported intent/action capabilities and limits rather than only numeric axis IDs.
- Give each materially different intent family its own scope/authority semantics. A velocity-to-attitude/direct-thrust transition is not an in-place interpretation toggle.
- Make scheme/intent changes transactional: neutralize, clear edges/integrators/filters, require controls neutral, publish new scheme revision/digest, and reacquire authority when scope changes.
- Keep simulation lifecycle actions outside vehicle flight intents (SIM-01).
- Preserve future 6-DoF translation/rotation intents and negotiated frame IDs without implementing orbital guidance in this issue.

## Guardrails / acceptance criteria

- [ ] Velocity and attitude/direct-thrust frames cannot be confused by type or scope.
- [ ] Arm/disarm/mode actions are explicit and one-shot, with acknowledgement/rejection surfaced.
- [ ] A held button/axis cannot fire under the new meaning after a scheme switch.
- [ ] A scope-changing switch cannot transmit until neutral and a fresh authority generation are established.
- [ ] Unsupported intent, unit, frame, or capability combinations fail closed.
- [ ] Profile/scheme revision and content digest are traceable through control evidence.
- [ ] Tests cover reconnect, link loss, stale generation, switch while non-neutral, and duplicate discrete actions.

[...]

Issue #108, opened 2026-07-15
INPUT-01: make pilotage-input the single native/WASM device-mapping engine


luofang34:
## Problem

Pilotage already has a versioned, layered `pilotage-input` profile model for device identity, axis calibration, deadzone, expo, inversion, and raw-button mapping. The browser does not consume it. `clients/web/main.js` maintains a second hard-coded controller/profile implementation, including controller matching, axis assignments, expo, deadzone, and arm/disarm key positions.

Two mapping engines will drift, and neither profile identity nor its exact content is currently strong enough for reproducible control/evidence.

## Scope

This issue owns only **physical device input → normalized logical controls**. Semantic flight schemes and typed vehicle actions are a separate child under LINK-01.

## Required change

- Compile the same Rust mapping engine for native and wasm/browser use.
- Load built-in and layered organization/user/vehicle/session profiles through the existing versioned profile model.
- Replace the browser's hard-coded controller calibration/mapping tables with the shared engine.
- Define deterministic device matching with an explicit generic fallback and visible selected-profile identity.
- Carry both a monotonic profile revision and a content digest on/control-adjacent to outgoing frames and evidence.
- Make profile replacement fail safe: neutralize output, reset button-edge/filter state, and require release/return-to-neutral before resuming.
- Reject unknown schema versions, unknown logical names, malformed calibration, NaN/Inf, and ambiguous profile matches.

“Pluggable” means declarative, versioned, validated profiles and registries first. Arbitrary runtime code plugins are out of scope for a flight profile; a future sandboxed extension must still emit through this validator.

## Guardrails / acceptance criteria

- [ ] Native and wasm golden vectors produce bit-identical logical axes/button edges for the same raw capture and profile.
- [ ] RadioMaster Pocket, Standard Gamepad/DualSense, keyboard, and generic fallback profiles are data, not JS branches.
- [ ] Layer precedence remains deterministic and is tested.
- [ ] Hot profile replacement emits neutral, clears stale edges/filter history, and cannot replay a held arm/mode action.
- [ ] The selected profile revision + digest are visible in session state/evidence.
- [ ] A structural/conformance guard prevents a second first-party browser mapping table from reappearing.

SIM / NOT FOR FLIGHT.

PR #170, opened 2026-07-19
GIM-03: LT+RStick gimbal quasimode on the consolidated #108/#109 input/control seam

luofang34:
## Review round 5: all four ingress/lifecycle claims verified TRUE and fixed (2026-07-22)

| # | Review claim | Verified | What landed | Evidence |
|---|---|---|---|---|
| 1 | `sim.lifecycle` must be command-only | TRUE — releasing it engaged link-loss, and with no intents the scope has no neutral demonstration, so the latch could never clear: PX4 permanently rejected the next reset | Any group whose members declare no intents, no legacy map, and no axes is COMMAND-ONLY: it never arms the frame-silence watchdog (`next_deadline()` stays `None`) and release/revoke paths skip link-loss engagement entirely. Two complete reset → release → re-lease cycles are pinned: each cycle's reset reaches the adapter, each lease lands strictly newer, no cycle engages the latch. | `the_reset_release_re_lease_cycle_stays_repeatable` |
| 2 | Expire overdue holders BEFORE processing messages | TRUE — a frame landing after the deadline but before the scheduled tick refreshed the expired deadline and resurrected authority | `expire_overdue` (offer expiry + holder-silence release) runs at the front of EVERY client message as well as the tick: a late frame finds its authority already revoked in the same call — link-loss engaged, neutralize emitted once, frame rejected `NoHolder`, and the deadline it missed is never refreshed. | `a_late_frame_after_the_deadline_cannot_resurrect_authority` |
| 3 | Harden datagram ingress | TRUE — `frame.session` was never checked, duplicates/reordering were admitted freely, and staleness subtracted the browser clock from the host clock | Ingress now (a) refuses a foreign `frame.session` (`SESSION_MISMATCH`) before any sender record is read, (b) enforces a wrap-aware strictly-advancing sequence watermark per group-pair per generation (`STALE_SEQUENCE`) BEFORE liveness refresh or recovery clearing — a fresh generation restarts the domain, and (c) judges staleness on correlated age (per-client minimum observed clock delta through `pilotage_timing::estimated_age`), never raw cross-clock subtraction. | `ingress_hardening.rs` (8 tests: session mismatch, duplicate + deadline pin, reorder, `u32::MAX → 0` wrap, fresh domain, late frame, 60 s skew correlation, typed-only default) |
| 4 | Typed-only must be the production default | TRUE — legacy payloads bypassed profile activation and carried uncorrelated Arm/Disarm edges on droppable datagrams | `SessionConfig.legacy_compatibility` defaults FALSE: a payload-bearing frame is refused `LEGACY_DISABLED` on every production host. Legacy translation survives only behind the explicit SIMULATION compatibility mode — `with_legacy_compatibility(true)` in test harnesses, `RuntimeOptions`/`PILOTAGE_LEGACY_COMPAT=1` (warn-logged) on the host for the legacy-only reference adapter loopback. | `typed_only_is_the_production_default`; host opt-in in `runtime/options.rs` |

**Evidence**: CTRL-INGRESS-006 added to `docs/control` (requirement, design, four implementation selectors, eight cases); CTRL-LIFE-005 extended with the command-only clause and the cycle case. All ten recorded results re-run against committed baseline `967ed8a1` with fresh git-blob source digests and sha256 artifact digests; the trace resolves bidirectionally (70 nodes each way), selectors resolve, the hard baselines-resolvable step passes locally, and **the review record remains PENDING until an independent review completes**. Pushed without waiting for CI per directive.

---

## Review round 4: all five items implemented and evidence re-baselined (2026-07-22)

| # | Review item | What landed | Evidence |
|---|---|---|---|
| 1 | Leased concrete member | A group lease binds the CONCRETE scope it was acquired for: the holder commands that member and only that member. A frame or action on the unleased sibling is a non-holder's — rejected even under the shared generation — and requesting the sibling while holding is denied `AlreadyHeld`: the switch goes release-first (neutralize, group latch engaged) and the sibling grant lands strictly newer in the one generation domain. `the_group_holder_may_drive_through_either_member_scope` is REPLACED by sibling-frame rejection, sibling-action rejection, and release-first handover tests. | `authority_group.rs` (6 tests) |
| 2 | Structural lifecycle profiles | The aviate adapter stores its session profile; `sim.lifecycle` exists only under `Simulation` — a `Physical` profile neither advertises nor executes lifecycle commands regardless of its uplink (advertisement gate + execution gate + a physical-profile test asserting no advertisement, whole-frame refusal, zero reset spawns). PX4 has no physical profile to construct (structurally sim-only); an engine test pins that an unadvertised lifecycle scope is structurally absent from the session — nothing to lease, nothing to command. | `a_physical_profile_neither_advertises_nor_accepts_lifecycle_commands`; `an_unadvertised_lifecycle_scope_is_structurally_absent` |
| 3 | Result guarantee | A correlated action the adapter left unanswered — an early rejection (link loss engaged, measurement unavailable, reset in progress, unknown scope) returns before per-action disposal — is answered by the actor with the rejection's reason: every valid command yields exactly ONE correlated result, never a browser timeout. Pinned at the actor (reason-agnostic synthesis) and over the REAL transport: a latched adapter refusing a bound Arm still answers id-for-id with `LinkLossEngaged` in the detail. | actor test + `reliable_actions_wire.rs` step 4 |
| 4 | Bounded anti-replay | The 64-entry cache is joined by a monotonic WATERMARK: an id at or behind the newest admitted one is permanently stale — refused, never executed — even after its result leaves the bounded cache, across the full u32 wrap (sound because one sender mints ids sequentially on one ordered stream). Tested: 80 actions past the bound then a replay of id 1 (answered, not executed), immediate replay, fingerprint mismatch, and the `u32::MAX → 1` wrap. | `an_evicted_id_stays_stale_and_never_executes_again`, `the_watermark_wraps_at_u32_without_stalling_fresh_ids` |
| 5 | Explicit acceptance events | The engine emits `SessionAction::ActivationAccepted` exactly once per validated announcement, and the actor's evidence line derives from that event alone — the state comparison is gone, so a rejected duplicate or foreign-session announcement (whose standing record still aliases the announced revision) can never be misreported as accepted. | `acceptance_is_an_explicit_event_and_a_rejected_duplicate_emits_none`; every announce-site test asserts the event |

**Evidence**: CTRL-GROUP-003, CTRL-CHAN-002, CTRL-BIND-001, and CTRL-LIFE-005 requirement texts updated; ten new verification cases joined the graph; all results re-recorded against committed baseline `8d1955c5` with fresh artifact/source digests. The trace resolves bidirectionally (53 nodes each way), selectors resolve, the hard baselines-resolvable CI step passes, and **the review record remains PENDING until an independent review completes**.

---

## Review round 3: all five rework items implemented, tested, and flown; gimbal video verified live (2026-07-22)

| # | Review item | What landed | Evidence |
|---|---|---|---|
| 1 | Actions on the reliable ordered stream | Typed actions never touch datagrams: each press travels once as a `ControlActionCommand` on the ordered session stream, bound to session + vehicle + scope + fencing generation + announced activation revision, with a REQUIRED nonzero correlation id. The engine validates every binding against its own records (forged session closes; non-holder / stale generation / unbound revision / zero id / unadvertised action each answer an explicit rejected result on the same stream). Dedup is by id PLUS immutable request fingerprint — a replay re-answers the cached result, a reused id with different content is refused. Datagram frames carrying typed actions are rejected whole (`ACTION_ON_DATAGRAM`). | 9 engine tests incl. **`a_delayed_arm_bound_to_a_superseded_generation_cannot_re_arm`**; actor fingerprint tests; a REAL-transport integration test (`reliable_actions_wire.rs`) driving bound→adapter→correlated-result, replay-re-answer, and stale-refusal over live WebTransport |
| 2 | Exclusive authority group | `ScopeDescriptor.authority_group`: scopes driving one actuator (aviate's `vehicle.motion` + `vehicle.motion.direct`) are ONE authority — leases, holder identity, fencing generations, the silence watchdog, the link-loss latch, and neutral recovery all key on the group; capability validation stays per concrete scope. Siblings can never be held simultaneously; a handover's regrant is strictly newer in the one generation domain; releasing one member engages the ONE group latch and the sibling's neutral demonstration clears it — no orphaned old-scope latch. The aviate adapter keys its flight latch by the group so a direct-scope frame can never fly around a motion brake. The browser also gained the missing revocation path found live: fenced rejections on the motion group drop local authority and re-lease instead of streaming rejected frames forever. | 4 engine tests (`authority_group.rs`); aviate suite; live swap flows on both backends |
| 3 | Keyboard as a layered profile + transactional sources | The keyboard resolves through the SAME layered registry as pads (reserved exact identity; wildcard pad overrides can never bleed into key bindings — pinned). The coordinator tracks the ACTIVE INPUT SOURCE: the announcement names the device actually driving — keyboard identity/revision/digest at boot and after any disconnect, the pad's after a selection transaction. Pad connect, disconnect (DOM event + poll backstop), and same-model replacement each run the neutral handover, cycle leases, advance the revision, and re-announce. | coordinator tests incl. `the_boot_source_is_the_keyboard_and_a_disconnect_returns_to_it`; shared golden groups (native + wasm) |
| 4 | SimReset off flight authority | `SIM_RESET` removed from both flight scopes and from the legacy velocity map (`reset_button` no longer exists). It lives solely on the new `sim.lifecycle` scope — separately leased, advertised ONLY by simulation adapters, never by a live/RF gateway. The browser leases the scope on demand for one press and releases when the result lands. Flight authority neither grants nor implies a reset. | engine test `a_sim_reset_needs_the_lifecycle_scopes_own_lease`; adapter lifecycle suites; live reset on px4-gz |
| 5 | End-to-end evidence | (a) Activation evidence is emitted only AFTER engine validation — a rejected announcement logs a warning, never a record. (b) `docs/control/`: requirement statements (CTRL-BIND-001, CTRL-CHAN-002, CTRL-GROUP-003, INPUT-SRC-004, CTRL-LIFE-005), an evidence graph tracing requirement → design → implementation → named `#[test]` cases → recorded results (7 suites, 33 cases) bound to committed baseline `2766d3fc`, resolving bidirectionally with selectors resolved; the evidence-gate workflow runs the slice with the same HARD baselines-resolvable step as LINK-04. (c) The interleavings run through wasm/wire → session host → adapter → reliable result: the browser-encoded fixture pair (`typed-frame-fixture.json` setpoint + action command halves) and the live-transport integration test. | `docs/control/evidence-graph.evg`; CI slice steps; `typed_wire_e2e.rs`; `reliable_actions_wire.rs` |

**Gimbal video**: verified live end to end on a default `cargo xtask sim --fc px4-gz` launch (the launcher enables the gimbal by default) — all three feeds paint and the gimbal camera physically slews under LT+RStick with the feed following. The investigation surfaced the real defect behind frozen-looking sessions: the browser had no handler for host watchdog revocations and would stream NoHolder-rejected frames forever; the self-healing reacquisition now covers both the motion group and the gimbal scope, and `FrameRejected` decodes its full addressing (scope, sequence, host generation) for diagnosability. On aviate the gimbal pane is black by design (no gimbal camera on that vehicle).

Round-3 live flights (px4-gz + aviate-gz, this head): announce→lease ordering with source-true device identity, stream-borne Arm acks with correlation ids, on-demand lifecycle reset, transactional DualSense hot-swap, and the aviate direct-scope round trip with group recovery.

Physical PX4 validation stays pending under #168. #108/#109 stay open per review direction.

---

## Review round 2: all six rework items implemented, tested, and flown (2026-07-22)

Every item from the review verdict is landed on this branch (commits `93c4b838`, `29f2b2f4`, `885c3499`, `2b26dd65`), gated green, and re-validated live on **both** SITL backends.

| # | Review item | What landed | Evidence |
|---|---|---|---|
| 1 | Profile traceability | `ProfileActivation` names the **composite** mapping (scheme + device profile id/revision/digest). The engine now **enforces**: an announcement naming a foreign session or regressing the monotonic revision closes the client; a typed frame whose `activation_revision` doesn't match the announcement is rejected `PROFILE_MISMATCH` before the gate (legacy loopback payloads exempt). The browser announces **before** the LeaseRequest on the same ordered stream and re-announces whenever the revision advances. | `tests/profile_binding.rs` (7 tests incl. u32 wrap); live: host log records `activation_revision=1→2→3` with device identity/digest on both backends, zero mismatch rejections in normal flight |
| 2 | Transactional device/profile changes + layered registry | `ControlCoordinator` (driven by BOTH the wasm shell and the native golden harness): device resolution goes per-layer `select_by_identity` (ambiguity fails closed) + `merge_layers` across built-in<org<user<vehicle<session (`add_device_profile` wasm/JS seam); an effective-mapping change re-opens the runtime's **neutral handover** — forced neutral, gimbal+motion lease cycle, map installs only at the boundary, edges reseeded through the new map, revision advances, re-announce precedes reacquisition. | `coordinator/tests.rs` (deflected-new-pad forced neutral & gated until regrant), reworked golden groups (transactional swap, session-layer override) run native+wasm; **live px4-gz**: fake DualSense hot-swap → re-announce rev 2 → release → regrant gen 3 → recovery ack gen 3, exactly one boundary-fenced frame |
| 3 | CTRL-01 scope model (Aviate) | The `fpv_mode` latch is **deleted**. Aviate advertises `vehicle.motion` (Velocity) **and** `vehicle.motion.direct` (AttitudeThrust, LOCAL_NED, tilt bound = the uplink's real 0.6 rad, new `max_yaw_rate` capability field). The gate rejects velocity-on-direct (`UNSUPPORTED_INTENT`) and bounds tilt via the shared quaternion→Euler. The browser FPV toggle is a **scope handover** (release → reacquire under a fresh generation); the client owns tilt scaling, client-integrated heading (seeded from declared telemetry heading), and the linear collective the adapter inverts bit-exactly. Level-at-mid-collective is the direct scope's neutral posture for link-loss recovery and reset clearance. | `direct_flight.rs` (4 tests: SET_ATTITUDE_TARGET on the wire, Euler round-trip, defensive rejection, attitude neutral), gate tests; **live aviate-gz**: both scopes advertised, toggle → `motion scope is now vehicle.motion.direct` granted fresh, attitude flight with zero rejections, toggle back runs the full neutral recovery on `vehicle.motion` (regrant gen 3 + recovery ack) |
| 4 | Capability negotiation | Packed repeated varint decode fixed (`repeatedVarints`, PACKED + unpacked); capability lookup keys on **vehicle id + scope**; every discrete action checks the advertisement before it is enqueued (suppressed loudly). `server-welcome-fixture.json` is a ServerWelcome **encoded by the real host** with multi-value packed frame/mode-target lists; `welcome_fixture.rs` pins the bytes, `wire.test.mjs` decodes them field-by-field. | Rust+JS fixture tests; **live px4-gz**: FPV press correctly suppressed (`not advertised`) since px4 advertises neither a direct scope nor a mode request |
| 5 | Delivery-complete discrete actions | Actions carry correlation ids (`action_id`): the client enqueues each press, retransmits it on **every** frame for its scope until the `ControlActionResult` echoes the id (1.5 s timeout reports failure loudly), and flips mode state **only on an accepted ack**. The host deduplicates per client (bounded cache), strips answered actions before adapter delivery, and re-answers replays — the vehicle executes each press exactly once. | `action-tracker.test.mjs` (8 tests: retransmit, dedup-preserving repeats, arm↔disarm cancellation, timeout, u32 wrap), actor `reliable_actions.rs` (4 tests); **live both backends**: `action result … id=N accepted` correlation |
| 6 | Artifact freshness closure | The wasm stamp's input set is the **derived** transitive workspace dependency closure of the wasm crates (`cargo tree`), unioned with a floor of the wasm crates, build script, `Cargo.toml`, and `Cargo.lock` — a protocol edit, lockfile pin, or profile change all rebuild. | closure regression test pins `crates/pilotage-protocol`, `crates/pilotage-input`, `Cargo.lock` in the stamp inputs |

Confirmed-good items from the verdict (exactly-one representation, typed adapter delivery, ambiguous-profile rejection, conditional gimbal video) are untouched.

**Round-2 live flights (2026-07-22, branch head `2b26dd65`):**
- **px4-gz**: welcome → capabilities → *composite* activation announcement → lease grant, strictly in that order on the reliable stream; typed climb + Arm acked with id; fake-DualSense hot-swap ran the full transaction (re-announce → release → fresh grant → host recovery ack) with the swap window fenced (single NoHolder rejection at the boundary, then clean).
- **aviate-gz**: two-scope advertisement decoded (`vehicle.motion intents=[1]`, `vehicle.motion.direct intents=[3]`); armed + climbed on velocity; FPV toggle handed over to the direct scope under a fresh generation (heading seeded from telemetry); attitude flight clean; toggling back re-acquired `vehicle.motion` through the full neutral-recovery handshake. Host recorded monotonic activations 1→2→3 with zero enforcement trips.

Physical PX4 validation stays pending under #168.

---

## #108 + #109 are complete inside this PR: one mapping engine, a typed-only control path, and the integration regressions fixed

GIM-03 (LT+RStick gimbal quasimode) ships on a **finished** INPUT-01/CTRL-01 seam: `pilotage-input` is the sole mapping engine on both sides of the wasm boundary, every control frame that reaches an adapter is a typed intent/action validated against the vehicle's own advertisement, and the whole chain is pinned by shared fixtures plus live flights on both SITL backends.

### Status matrix — every stage, per backend

| Stage | Mechanism | Evidence | reference | px4-gz | aviate-gz |
|---|---|---|---|---|---|
| Device → logical mapping | `pilotage-input` DeviceProfile data (keyboard keys, DualSense, RadioMaster Pocket AETR reroute, generic wildcard) compiled in the wasm device stage; **no JS table survives** | `device/tests.rs`, shared golden vectors, structural guard over **all** `clients/web/*.js` | n/a (host-side) | ✅ live: `gamepad mapped (exact): Sony DualSense` | ✅ same client path |
| Device identity | `Gamepad.id` → `select_by_identity` (Chromium+Firefox forms); ambiguity **refuses the pad** (no map, empty samples) | `select.rs` tests (AmbiguousExact/Wildcard), `device/tests.rs`, golden vector group 6 | n/a | ✅ live exact match | ✅ same |
| Golden parity | ONE fixture (`golden-vectors.json`) executed by the native harness AND the compiled wasm | `golden.rs` + `control-runtime.test.mjs` | ✅ | ✅ (same wasm flown) | ✅ |
| Profile traceability | `ProfileActivation` (envelope 16): id + doc revision + monotonic activation revision + SHA-256 digest; frames carry `activation_revision` (field 11); host records per session + structured log | engine test `profile_activation_is_recorded_for_evidence`; **live: viewer digest line == host log digest** (`29e2e20af978…`) on both backends | ✅ engine test | ✅ live | ✅ live |
| Typed emit (browser) | plan × **advertised** envelope → `VelocityIntent`/`GimbalRateIntent` + typed actions; fail-closed without an advertisement | `typed-command.test.mjs` (keyboard **and** gamepad through real wasm, byte-pinned fixture) | n/a | ✅ live | ✅ live |
| Capability negotiation | Adapters advertise real `IntentCapability` (family/frames/limits incl. `max_vertical`) + `ActionCapability` (mode targets); projected into `ServerWelcome` | `welcome_advertises_the_typed_capability`; **live: `vehicle.motion intents=[1] actions=3`, `vehicle.gimbal intents=[5] actions=1` (px4), gimbal scope absent (aviate)** | ✅ unit | ✅ live | ✅ live |
| Command gate | Exactly-one representation (EMPTY/DUAL), family+frame+limit checks, action/mode-target checks, duplicate + Arm∧Disarm conflicts — **before** adapter delivery | `command_gate/tests.rs` rejection matrix (14 tests); convert-layer rejects non-finite, non-unit quaternions, thrust∉[0,1], targetless mode requests | ✅ | ✅ | ✅ |
| Legacy boundary | ONE translator (`command_gate::translate_legacy`) driven by adapter-declared `LegacyCommandMap` data; structurally-total translation rejects partial payloads (`PARTIAL_COMMAND`); adapters **reject** any payload frame | gate tests; adapter typed-only rejection tests ×4; `loopback.rs` drives a legacy frame through the real WebTransport host into the typed skiff | ✅ loopback | ✅ (browser is typed; boundary idle) | ✅ same |
| Adapter consumption | Typed velocity/gimbal-rate ÷ the same advertised limits (full envelope == full stick); typed Arm/Disarm/SimReset/ModeRequest{target}/GimbalRecenter | pinned skiff golden trajectory reproduces **exactly**; per-adapter suites (36 px4 / 53 aviate / 24 gazebo / 30 reference) | ✅ bit-exact | ✅ live OFFBOARD+arm, gimbal claim + slew + recenter | ✅ live `takeoff: climb input opens the setpoint stream` |
| Action acknowledgement | `ApplyOutcome.action_results` → `ControlActionResult` (envelope 17) to the sender; viewer logs each | actor tests; **live: `action result [vehicle.motion seq=120] action=1 accepted`, `[vehicle.gimbal seq=287] action=4 accepted` (px4); Arm accepted (aviate)** | ✅ | ✅ live | ✅ live |
| E2E chain | `typed-frame-fixture.json`: browser half asserts the wire bytes byte-for-byte from real wasm; Rust half decodes the SAME bytes through welcome→lease→gate→adapter boundary | `typed-command.test.mjs` + `typed_wire_e2e.rs` | ✅ | ✅ (chain + live) | ✅ |
| Profile-switch safety | Transactional handover + motion-lease cycling + neutral activation + host-confirmed recovery (pre-existing rounds 5–12); device **reselect reseeds edge baselines** (pad-swap can't fire a held arm) | runtime/motion tests, golden groups 4–6, `uni-stream`/`motion-lease` node tests | ✅ | ✅ | ✅ |
| Recovery gate (typed) | Neutral activation is a TYPED demonstration (`intent_satisfies_neutral_activation` vs advertised limits, no actions); actions-only frames prove nothing | recovery tests incl. rewritten partial-coverage + actions-only cases | ✅ | ✅ (per-scope link-loss unchanged) | ✅ |
| Preflight | Existence checks → **content stamps** (working-tree source hashes + toolchain versions) for viewer wasm and gz sidecar | xtask stamp tests; **live: both rebuilt on this branch's source changes, then skipped** | — | ✅ live | ✅ live |
| Gimbal video | Bridge subscribes + adapter advertises source 2 **only when the vehicle is configured with a gimbal** | positive+negative unit tests (`bridge_config`, `advertised_video_sources`); live: 3 feeds (px4 w/ gimbal), gimbal pane absent for aviate | n/a | ✅ | ✅ (negative) |

### Live validation (2026-07-22, this branch head)
- **px4-gz** (`PILOTAGE_PX4_GIMBAL=1`): capability negotiation → DualSense exact profile → `ProfileActivation` digest matched host-side → typed **Arm** → `offboard arm sequence: OFFBOARD and arm commanded`, telemetry ARMED → typed velocity climb → LT+RStick **gimbal-rate** slew (`Configured primary gimbal control … to 255/190`) → typed **GimbalRecenter** (`gimbal neutral commanded`) with explicit accepted results for both actions. FPV pane paints live sim video; PFD/HSI live.
- **aviate-gz**: velocity-only advertisement, **no gimbal scope**; gimbal lease denied with flight unaffected; typed Arm accepted; typed climb opens the FC setpoint stream.
- The live flights also caught (and this PR fixes, with regression tests): the wire projection dropping the advertisement, the activation announcement racing the session writer, and action results missing from the post-bootstrap reader — exactly what the fail-closed design should surface.

### Gates
`cargo fmt --check` · `clippy --all-targets -D warnings` (0) · `cargo test --workspace --all-targets` (0 failures, 40+ suites) · `check-structure` OK · all 31 `clients/web/*.test.mjs` including the real-Chromium viewer-boot and control-runtime browser tests · CI (quality-gates + evidence-trace-gate) green on the branch.

### Explicitly deferred (tracked, not silent)
- **Physical PX4 validation stays pending under #168** — it was NOT performed here. The earlier SITL fault-injection acceptance (PX4 `6120aa53df…`, host stop provably dropped, PX4's own 2 s timeout as sole failsafe) remains recorded in ADR-0008/`px4_gz.rs`.
- ~~Aviate FPV/direct flight still consumes velocity-family intents after `ModeRequest{FpvDirect}`~~ — **done in round 2**: direct flight is `vehicle.motion.direct` (AttitudeThrust) under its own lease; the mode latch is gone.
- The RadioMaster Pocket browser profile carries its HID-capture-derived AETR routing with an explicit needs-physical-verification note.

Gimbal feature summary, the round-5…12 link-loss/recovery record, and the SITL fault-injection acceptance are unchanged below.

---

## PX4-GZ acceptance flown — validated SHA recorded (round-12 blocker 3 closed)

The discriminating fallback test was actually flown on the `px4-gz` backend (PX4 `6120aa53df874021639e2413a4cdecf8df8e355a`, `v1.18.0-beta1-110-g6120aa53df`) with the fault injection `PILOTAGE_PX4_DROP_GIMBAL_STOP=1`:

- PX4 accepted Pilotage's primary-gimbal-control claim: `[gimbal] Configured primary gimbal control ... to 255/190`.
- On holder disconnect, reproducibly across three disconnects, the session host logged `holder lost; engaging link-loss policy scope="vehicle.gimbal" ... HolderDisconnect` immediately followed by **`gimbal link-loss stop DROPPED (fault injection); relying on PX4's own timeout`** — the host provably sends **no** stop, so PX4's own 2 s timeout (`output.cpp check_and_handle_setpoint_timeout`, code-verified in that exact checkout) is the sole failsafe. This is the discriminating property the review asked for — Pilotage's queued stop can no longer mask the test.
- Per-scope link-loss (gimbal + motion independent) and the `LinkLossCleared` recovery ack (`recovery confirmed on generation=5`) were also verified live in the same session.

The exact validated SHA is recorded (not a range) in the `px4-gz` backend doc and ADR-0008. The physical rate-zeroing trace at T+2 s is left to **#168** (PX4-in-loop evidence).

## Round-12 review fixes — trait postcondition doc, discriminating PX4 fallback test, log de-storm

1. The **asymmetric latch postcondition is now on `VehicleAdapter::set_link_loss_policy` itself** (a "# Latch postcondition" section), not only ADR-0008: an engage records the latch even on `Err`; a clear removes it only on `Ok`; a scope is un-suppressed only by an accepted clear.

2. Removed the last two false **"verified zero-rate"** claims in PX4 module headers (`adapter/link_loss.rs`, `gimbal_link_loss_tests.rs`) — both now say "queued best-effort".

3. The manual PX4-fallback acceptance is now **discriminating**. A plain flight passed on the host's own stop and never exercised PX4's timeout. Added **fault injection** — `PILOTAGE_PX4_DROP_GIMBAL_STOP=1` makes `queue_link_loss_stop` drop the host's zero-rate entirely (a unit test proves it sends **nothing**: no claim, no rate), so PX4's own ~2 s gimbal-manager setpoint-timeout is the **sole** failsafe under test. The `px4-gz` bring-up doc now gives the discriminating procedure and a slot to record the **exact validated PX4 SHA** (not "at or after"); ADR-0008 evidence updated. It stays manual — **#168 tracks the PX4-in-loop evidence**; no automated in-loop test runs in CI.

4. Retried clears no longer emit `debug!` at 100 Hz. Logging is **transition-only**: `error!` + count on the first refusal, `info!` on the recovery (a retry that takes), and silence for still-refused retries.

`gimbal.rs` unit tests moved to a sibling `gimbal/tests.rs` to stay under the 500-line file gate.

The reviewer's recommendation to extract the link-loss lifecycle into a dedicated prerequisite PR still stands (noted for the author's decision; kept in #170 per the standing directive). #170 stays Draft until #108 and #109 land; #168 stays open for PX4-in-loop evidence.

## Round-11 review fixes — fail-closed clear postcondition, de-stormed retry, honest fallback

1. **Failed-clear postcondition made explicit and fail-closed.** ADR-0008 no longer says the latch is recorded "regardless of success" for a clear — the postcondition is **asymmetric**: an **engage** records the latch even when the actuation is refused (a fenced scope must never be left drivable), and a **clear removes it ONLY on `Ok`** (a refused clear keeps the scope suppressed). The PX4 adapter is restructured so the clear drops the latch on the success path, never before an actuation that could fail. A new **stateful adapter test** proves a failed clear still rejects control (`apply_control` keeps returning `LinkLossEngaged` until a clear the adapter accepts).

2. **The tick-driven clear retry no longer storms.** The engine re-emits the clear every 10 ms while a scope is pending; the actor previously counted and `error!`-logged each one (a 100 Hz storm). `ClearLinkLoss` now carries a `retry` flag (`false` for the neutral-activation attempt, `true` for a tick re-emission); the actor counts and error-logs **only the first refusal** and debug-traces retries. Test: 50 refused retries add exactly one fault.

**Cleanup:**
- Removed the false claim that the host 300 ms cutoff re-sends after a gimbal link-loss stop — `queue_link_loss_stop` sets `streaming = false`, so `maintain()` no-ops. The declared independent backstop is **PX4's own gimbal-manager setpoint-timeout** (zeroes a nonzero rate after ~2 s, `src/modules/gimbal/output.cpp`, behavior pinned at commit `841bb40`), documented at the call site, in ADR-0008, and as a **manual acceptance criterion** of the `cargo xtask sim px4-gz` bring-up (no automated PX4-in-the-loop test runs in CI).
- ADR evidence text: "verified" gimbal zero-rate → "queued best-effort".

The reviewer's recommendation to extract the link-loss lifecycle into a dedicated prerequisite PR still stands and is noted for the author's decision; the fixes are landed here per the standing directive. #170 stays Draft pending #108/#109.

## Round-10 review fixes — engine-owned clear lifecycle (no stale clear across a generation) + honest gimbal semantics

Two safety-critical correctness gaps.

1. **A stale failed-clear could cross an authority generation change.** The pending-clear retry lived in the engine actor, which cannot see holder changes, so a committed handover or emergency override — which advance the generation with **no** re-engagement — could let a stale clear un-latch the scope for the new holder *without* its neutral activation, violating ADR-0010.

   Recovery is now an **engine-owned state machine: `Engaged → ClearPending(generation) → Cleared`**. A neutral activation moves the scope to `ClearPending` and emits `ClearLinkLoss`, but the engagement marker is **not removed until the driver reports the adapter confirmed the clear** (`SessionEngine::confirm_link_loss_cleared`, generation-gated). `handle_tick` re-emits the clear for a still-pending scope (so a refused clear retries with no actor-held state), and **any holder-installing effect (grant / committed handover / emergency override) invalidates a pending clear**, reverting it to `Engaged` — so a stale clear can never cross a generation, and the new holder must re-demonstrate neutral activation.

   Tests: `LinkLossState` state-machine unit tests (invalidate reverts to Engaged; a stale-generation confirm is rejected; a fresh generation is required); **engine handover + emergency-override race tests** (a pending clear is invalidated and no stale retry crosses the generation); actor integration tests driving the real recovery flow (a confirmed clear acks once; a refused clear is retried by the engine and acks once when it takes).

2. **The PX4 gimbal stop is queued, not FC-"verified".** `link_loss_stop` reported success on local channel acceptance, with no `CONFIGURE` acknowledgement and a claim `claim_if_due` could skip. Renamed **`queue_link_loss_stop`**, it now **force-re-asserts** primary control and documents that success means the zero-rate reached the FC lanes (**best-effort**), not that the FC confirmed a stop — the **declared safety net** is the FC's own setpoint-timeout failsafe plus the host's stale-demand cutoff. ADR-0008 amended to say link-loss actuation is best-effort at the link boundary (an adapter that *can* confirm should stay engaged/faulted until the vehicle confirms).

**Cleanup:** `uni-stream.test.mjs` now synchronizes deterministically via a pull-driven `ReadableStream` (checkpoints in the `pull` callback) rather than `setTimeout(0)`.

**Follow-ups (acknowledged, not in this round):** a dedicated link-loss slice in `docs/link/evidence-graph.evg` is deliberately deferred until the behavior settled (this round) — the evidence gate currently covers the LINK-04 source-role slice, and the new ADR MUSTs are traced by the tests above pending that slice. The reviewer's suggestion to extract the per-scope link-loss / reliable-recovery work into its own prerequisite PR is noted for the author's decision; it is kept here per the standing directive to land the fixes in #170. #170 stays Draft pending #108/#109.

## Round-9 review fixes — gimbal actually stops, a refused clear retries, ADRs reconciled, production reader under test

Four blockers on the per-scope link-loss work, all resolved:

1. **Gimbal link-loss now ENACTS `Neutralize`, it does not merely latch.** Engaging `vehicle.gimbal` drives a **verified zero-rate** gimbal setpoint immediately — re-asserting primary control first (PX4 drops a rate from a non-primary sender) — so an in-flight slew stops now instead of coasting until the far slower 300 ms stale-demand cutoff, and without ever touching flight. A refused send is a typed `ChannelRejected` fault; the latch engages regardless.

2. **A refused clear no longer strands recovery.** The engine un-engages the scope before the fallible adapter clear, so a failed clear previously left the scope both un-engaged (no later neutral frame could retry) and un-acked (client stuck neutralizing forever). The engine actor now holds a **pending-clear** and retries it every tick until the adapter accepts it, then broadcasts the `LinkLossCleared` ack **exactly once**; retries are not re-counted as faults. New test: `failure → later success → exactly one ack`.

3. **The accepted architecture docs are reconciled.** ADR-0008 carried the old vehicle-wide `set_link_loss_policy(vehicle, policy)`; it now documents the per-scope, fallible signature with an **Amendments** section, and ADR-0010's *Link loss* section states that engage/clear/enact are **per scope** (the supported-actions menu and policy *selection* stay per vehicle). ADR-0003 updated for consistency. (The link subsystem's evidence graph does not reference link-loss, so it needed no change; the amendment points at the tests that are the evidence.)

4. **The browser regression now exercises the PRODUCTION reader, not just a helper.** Extracted `readUniStream` (the core of `readOneUniStream`) and `applyMotionRecovery` (the recovery-state transition) into testable modules. The new `uni-stream.test.mjs` drives the **real reader** over a **real `ReadableStream`** with the `0x01` authority kind tag, a `LinkLossCleared` arriving **fragmented on an OPEN (never-closing) stream**, and the **real recovery transition** together — asserting `motionRecovered` flips **live, before any close**, and **only** for the matching generation. Wired into CI.

To stay under the 500-line file gate, px4's `set_link_loss_policy` body moved to `adapter/link_loss.rs` and the new gimbal link-loss tests to `adapter/gimbal_link_loss_tests.rs`.

## Round-8 review fix — link-loss is scope-specific, and the recovery ack fires only after a confirmed clear

Two related design flaws the round-7 review flagged are now resolved end to end.

**Link-loss is per-scope, not vehicle-wide.** Releasing or losing `vehicle.gimbal` previously engaged the vehicle-wide flight link-loss policy, so dropping the gimbal could neutralize (brake) the aircraft. Now:

- Every adapter keeps a per-scope latch (`BTreeMap<ScopeId, LinkLossPolicy>`). `apply_control` suppresses a frame only while **its** scope is engaged, so a gimbal-scope failsafe latches gimbal frames without ever touching motion, and vice versa. `set_link_loss_policy(vehicle, scope, policy)` neutralizes flight only for the motion scope — a gimbal-scope engage never actuates the FC. Threaded through all four production adapters (gazebo, reference-headless, aviate, px4) and the `VehicleAdapter` trait.
- The session engine emits `EngageLinkLoss`/`ClearLinkLoss` **per scope** (every lost scope engages independently on the uncapped safety lane), and `ClearLinkLoss` now carries the scope + recovered generation.

**The `LinkLossCleared` ack fires only after the adapter confirms the clear.** The round-7 ack was broadcast from the pure engine's `maybe_activate_recovery`, which cannot see the fallible adapter result — so it acked even when the clear failed. The ack now lives in the engine actor (`hosts/session-host`) and is broadcast **only** after `set_link_loss_policy(scope, None)` returns `Ok`. A refused clear leaves the vehicle neutralized, emits no ack, and is counted as a fail-closed fault, so the recovering client keeps neutralizing rather than resuming on a clear the vehicle never enacted.

**Tests**

- **px4 two-scope suppression** — engaging the gimbal scope leaves motion flying and never sends a neutral to the FC; engaging the motion scope suppresses motion while the gimbal keeps pointing (`adapters/px4/.../gimbal_tests.rs`).
- **Actor ack gating** — a confirmed clear broadcasts exactly one `LinkLossCleared`; an adapter-refused clear broadcasts nothing and increments the fail-closed fault counter (`engine_actor/tests.rs`).
- **Engine per-scope recovery** — a neutral frame on an unrelated scope cannot clear another scope's engagement; only a fresh grant + neutral demonstration on the lost scope recovers it (`tests/recovery.rs`).
- **Disconnect fan-out** — a client holding N scopes engages N independent link-loss policies on the safety lane, above the per-call broadcast cap (`tests/disconnect.rs`).

**Draft** — the LT+RStick gimbal quasimode, built on the shared Rust/WASM control runtime. This PR is the focused GIM-03 slice of #167; it intentionally leaves #167 open and does not attempt the broader input/mapping refactor, which is decomposed out (see Scope).

## What this does

Holding **LT** redirects the right stick to `vehicle.gimbal` LOS-rate frames at the control rate. While LT is held the right stick and LT read neutral to every flight scheme, so flight is masked cleanly; **R3** fires a one-shot recenter. The gimbal holds its own scope lease/generation/sequence; a denied lease is non-fatal (flight continues, gimbal frames simply aren't emitted). The mapping runs through the shared Rust → WASM control runtime (`clients/web-control`), not a JS mapping table — the JS shell only marshals pad/keyboard samples in and a typed plan out.

## Round-7 review fixes — terminal denial, host-acknowledged fence, and a RELIABLE recovery ack

This round makes the recovery genuinely reliable rather than best-effort, and closes two correctness gaps:

- **Terminal denial** — a denied motion reacquire now ignores every later response until a new session; a stale grant can no longer restore authority after a denial.
- **Host-acknowledged fence + u64 wrap** — the release fence uses the host's `LeaseReleased.generation` (the protocol's current fencing generation), and generation ordering is u64-modular (`u64::MAX → 0` is an advance; the ambiguous half-window fails closed).
- **Reliable recovery acknowledgement (new wire message)** — the client no longer resumes on a best-effort neutral burst. It stays neutralizing (retransmitting a neutral activation frame under the fresh generation every tick) until the host **confirms** it cleared the vehicle's link-loss latch, via a new reliable `LinkLossCleared { vehicle, scope, generation }` message (envelope tag 15) broadcast from the session engine's `maybe_activate_recovery` (safety-critical, undroppable, on the reliable authority stream). The client correlates by vehicle/scope/fresh-generation before resuming, so no dropped datagram or stale notice can desync the client and host.

This touches the session engine and wire protocol (`schemas/`, `pilotage-protocol`, `pilotage-session`, `session-host`) in addition to the client — the host-coupled recovery protocol the earlier rounds flagged as the remaining hardening.

## Round-6 review fix — recovery completes through a neutral activation burst

Round-5 gated the stale generation but resumed live the instant the lease regranted — so the first post-grant frame could be a live deflection, which never satisfies the host's neutral-activation recovery (`link_loss.rs` `payload_satisfies_neutral_activation`), leaving the host braking while the client thought control had resumed. The handshake is now complete:

- **Neutral activation burst** — after a fresh grant the runtime enters a `Neutralizing` phase: live output stays gated until the operator's controls read neutral, then it transmits a burst of neutral frames under the fresh generation (the activation the host needs to clear its link-loss brake) before live resumes. Because the operator is already neutral at resume, no command jumps.
- **Generation fence** — the shell accepts a regrant only when its generation strictly exceeds the released fence, so a stale/replayed grant can never resume control. The fence and the transition are a pure, unit-tested module (`clients/web/motion-lease.js`, `motion-lease.test.mjs`).
- **Terminal denial** — a denied reacquire stops all re-requests (a `Denied` phase) until a fresh session re-primes on reconnect.
- The motion-authority state machine moved to `clients/web-control/src/runtime/authority.rs`.

**Boundary (tracked for #109/host):** a host-*acknowledged* recovery — the client waiting on a session-stream recovery notice instead of best-effort neutral datagrams — needs a host-protocol addition, since `ClearLinkLoss` is an internal session action with no client-facing wire message today. The neutral burst is the correct best-effort client behavior; the reliable ack is the remaining hardening for flight use.

## Round-5 review fix — motion authority is reacquired before a remapped scheme transmits

The round-4 handover cycled the motion lease but emitted the reacquire request *beside* the newly-mapped motion frame, so the new scheme could publish a flight command on the released, stale generation before the host regranted. A live profile switch now runs a real motion-authority handshake in the runtime:

`release → wait for the session to reflect the release → request → gate all motion output until the host regrants on a fresh generation → resume`.

- **Motion frames are gated** (`motion: None`) for every phase but `Held`, so no flight command — not even a live stick — publishes on the released generation.
- **Release strictly precedes request**: the request is not emitted until the session confirms the release; a dropped release/request is re-emitted after a retry window instead of wedging the handshake.
- **The shell handles the motion `LeaseResponse` post-bootstrap**, installs the fresh generation, and restarts the sequence *before* any frame rides it; a release ack drops local authority. Arm/disarm gate with motion (baselines still advance, so nothing fires on resume).
- **The tick is genuinely allocation-free** — `Frame` is fixed-capacity (bounded axis/edge arrays, `Copy`) instead of building `Vec`s each tick, so the `wasm.rs` claim is now true.

## Round-4 review fixes

- **Fail-safe LT masking** — LT suppresses flight only when a gimbal lease is actually held. One effective condition (`held && active_gimbal`) drives flight masking, gimbal routing, and the HUD capture flag together, so a denied lease no longer silently zeroes throttle and right-stick.
- **Transactional profile handover** — a profile switch waits for the *union* of the active and candidate flight/gimbal controls to be neutral before installing (round-5 then fences and reacquires the motion generation, above).
- **Device revision on the wire** — control frames carry the compiled device-profile revision (ADR-0007/0009), kept distinct from the activation epoch.
- **Profile validation** — compilation rejects out-of-range axis/button indices (past the 8-axis / 24-button buffers), ambiguous or cross-action bindings, and unreasonable deadzone/expo values.
- **Session priming** — a fresh session generation seeds the runtime's edge baselines, so a control held across a reconnect fires no phantom edge.

## Guardrails

- Native golden vectors (`clients/web-control/src/golden.rs`) and the JS golden suite drive the **same** vectors through the real wasm, so a native/wasm divergence reddens one of them.
- Native ordering/gating/retry tests reproduce the round-5 defect (a live command on the stale generation) and prove it gated out; a wasm golden vector drives the same gate through the ABI.
- A `vehicle.gimbal` ControlFrame round-trips through decode in `wire.test.mjs` (scope, profile revision, axes, recenter edge).
- The hold-across-reconnect no-edge behavior is exercised through the real wasm runtime in the CI golden vectors (not a JS helper).
- Viewer-boot, control-browser-boot, and structural contract tests gate the JS shell.

## Scope (decomposition)

Per the review's recommendation, the broader input/mapping work is tracked separately and **not** built here:

- **#108 (INPUT-01)** — extracting the shared `pilotage-input` device-profile schema (RadioMaster/device matching, registry layers, deterministic fallback) so it isn't duplicated between the Rust runtime and the JS keyboard table.
- **#109 (CTRL-01)** — the logical-control → typed-intent seam that #162 wants underneath this.

#108, #109, #162, and #167 stay open. This PR remains a Draft prototype of the GIM-03 behavior on top of the runtime, pending that seam.


luofang34:
Review repairs pushed. The CI-red boot failure and the lease bugs are fixed:

1. **Boot failure (cross-lane leak).** The committed `main.js` had picked up two imports from a concurrent lane (`length-delimited-stream.js`, `latest-render-queue.js`) that don't exist on this branch's base — that's what failed the boot-contract CI. Removed. The trackpad wheel handler that depended on `#mainSlot` (which only exists in #173) is **entirely removed from this PR** and parked on a separate lane stacked on the layout branch, so this PR is once again independently revertible. Verified booting from its own tree in an isolated worktree (all local imports resolve; gimbal-input + wire suites pass).\n2. **BigInt vehicle identity.** `Number(m.vehicleId) === VEHICLE_ID` was always false (Number vs BigInt), so the LeaseReleased matcher never acknowledged. Now compares BigInt to BigInt.\n3. **Vehicle-checked gimbal responses.** Gimbal `LeaseResponse` routing now checks vehicle **and** scope, so another vehicle's response can't install our generation.\n4. **Held-lease liveness.** The gimbal lease is managed from the control loop: rover mode **releases** it (a held-but-silent scope would trip the host's per-vehicle holder-silence watchdog), flight modes re-request on entry, and a denied scope is never re-requested for the session.\n\nVerified live against both FCs: Aviate correctly **denies** the gimbal lease (no gimbal) and logs "flight control is unaffected" while motion control + arm work; px4-gz grants it and LT+stick drives the camera. CI green.

luofang34:
Converted to **draft** per review. The two concrete bugs are fixed (5be3889):

- **R3 false-edge:** `maybeSendGimbalFrame` advanced the R3 baseline only *after* its early return, so an R3 held across a lease grant, reconnect, or rover→flight switch read as a fresh rising edge (false recenter) when the path re-activated. The baseline now advances unconditionally inside a pure `gimbalResetEdge` helper; the edge fires only on a genuine press while active.
- **Untested lease state machine:** the request/release/none decision moved into a pure `gimbalLeasePlan` helper with behavioral tests (flight holds + re-requests, rover releases, denied never re-requests, R3-held-across-reactivation fires no edge).

**The architectural objection stands and is the owner's call.** #162 wants the quasimode to flow through the shared Rust/WASM device-mapping engine (#108 INPUT-01) and the typed control-scheme path (#109 CTRL-01), not a second JS mapping table. This PR is a working, now-correct JS *prototype*. Two options:

1. **Keep it a non-closing draft** as the reference behavior, and rebuild the mapping on the #108/#109 seam once those land (this PR then informs, but does not close, #167).
2. **Rebuild now** on #108/#109 — larger, blocks on those issues.

I've dropped the "Closes #167" so #167 stays open either way. Which direction do you want?

(The trackpad two-finger steering is parked on a separate lane stacked on this one, out of scope here.)

PR #183, opened 2026-07-22
Report gated arm/disarm presses instead of silently consuming them

### LINKED ISSUE #180: Arm/disarm presses vanish silently whenever motion output is gated
## Symptom

Reported flow: after a disarm, pressing Enter to re-arm does nothing. The gimbal still responds; flight does not. There is no overlay change, no log line, no `ControlActionCommand` on the wire — the press vanishes.

Reproduced live against `cargo xtask sim --fc px4-gz` at main `73a746cb`: in a session whose motion authority was not live (see states below), a real Chromium Enter keypress produced zero observable output while the gimbal lease stayed granted and gimbal frames kept flowing.

## Root cause

An arm/disarm press only becomes a `ControlActionCommand` when the motion-authority state machine says live output is allowed, and every link in that chain drops the press *silently*:

- `clients/web-control/src/runtime.rs` (`evaluate_active`): `arm: live && outcome.arm` — the typed edge is masked unless `advance_motion_authority` returned `MotionOutput::Live`. The edge baseline still advances (`edge_fired` updates `prev_arm`), so the press is consumed, not deferred.
- `clients/web-control/src/runtime/authority.rs` (`advance_motion_authority`): `Live` requires `MotionPhase::Held` *and* `session.motion_granted`. `Releasing`, `Reacquiring`, `Neutralizing`, `Denied`, and plain not-granted all gate.
- `clients/web/main.js` control loop: the action request itself lives inside `sendMotionFrame` (`plan.arm && requestAction(...)`), and `sendMotionFrame` is only called when `plan.motion` is non-null. A gated tick therefore sends nothing and logs nothing. The loud "action not advertised" and "no session stream" paths in `requestAction` are never reached.

The gating itself is correct fail-closed behavior — no arm may ride absent authority. The defect is that a safety-relevant operator press is swallowed with **zero feedback**, which the operator cannot distinguish from a dead keyboard.

States that produce exactly "gimbal works, flight and Enter dead", all observed live:

1. **Telemetry-only connect.** Autoconnect while `document.hasFocus()` is false: the lease probe `() => manual && gate.mayPublish()` (`clients/web/connect-authority.js`, `clients/web/control-gate.js`) skips the motion `LeaseRequest` entirely. The only trace is one line in the *collapsed* log dock ("no control lease granted; viewer is telemetry/video only"); the overlay keeps saying "connected — awaiting telemetry".
2. **Watchdog churn in a throttled tab.** In an occluded window (`visibilityState === "hidden"`), a 33 ms `setTimeout` measured 576–1000 ms; the 30 Hz loop runs at ~1 Hz, the host silence watchdog revokes between ticks, and the log cycles `control frame rejected (reason 2)` → `host fenced our motion authority; reacquiring` → grant → `LinkLossCleared` → revoked again (observed motion generation climbing 9 → 111 in minutes). Any Enter pressed during the gated portions of that cycle vanishes.
3. **Any post-revocation reacquire/neutralize window**, including after a profile/device handover.

## Why the current guard misses it

`clients/web-control` golden tests assert `plan.arm` is false while gated — the masking is tested as a *safety* property, but no test asserts the press produces an operator-visible diagnostic. No browser test dispatches a key while authority is degraded.

## Suggested fix

Feedback only — do not weaken the gate:

1. Have the runtime report a suppressed edge instead of erasing it: e.g. extend the plan flags with `arm_suppressed` / `disarm_suppressed` (edge fired while `!live`), or return the raw edges alongside `live` and let the shell decide.
2. In `main.js`, on a suppressed edge: log a specific line (`arm press suppressed: no motion authority (<phase/reason>); press Connect to resume control`) and flash the overlay. This mirrors the existing loud paths ("action X not advertised", "no session stream").
3. Surface the motion-authority state persistently in the control readout (the `els.gamepad` line) so "gated" is visible before the press, e.g. `| motion: live` vs `| motion: gated (no lease)`.
4. Bonus for state 1: when the lease probe skips control at connect time, reflect it in the overlay (not only the collapsed log), since this state otherwise looks identical to a healthy telemetry wait.

## Regression guard

- web-control unit/golden test: a tick with an arm rising edge under each non-live phase yields the suppressed-edge diagnostic, and the edge baseline still advances (held key does not re-fire).
- Browser test (extend `viewer-boot` or `control-runtime.browser.test.mjs` pattern): boot against a host stub that grants no motion lease, dispatch a trusted Enter via CDP, assert (a) no `ControlActionCommand` is written and (b) the suppression line appears in the log.

## Acceptance criteria

- Every arm/disarm press in a connected viewer produces exactly one visible response within one tick: either `arm command sent` or a suppression line naming the reason.
- The gating behavior itself (no action without live authority) is unchanged.

Related: #178 (result contract), #160 (typed emergency disarm), #146 (reconnect never resumes control). Companion issues: PX4 enactment feedback, and the autolaunch lifecycle issue (states 1–2 arise most often from the auto-opened tab).

luofang34:
Fixes #180. An arm/disarm edge that fired while motion output was gated was consumed with zero feedback — indistinguishable from a dead key.

- Runtime: `ControlPlan.arm_suppressed` / `disarm_suppressed` report a consumed-but-gated edge; wasm flags bits 6/7; shell decodes them.
- Viewer: each suppressed press logs its gate reason and flashes the overlay; the persistent readout now shows `motion: streaming|gated`; a lease-less session states itself on the overlay, not only the collapsed log.
- The gate itself is unchanged: no action ever rides absent authority.

Guards: new shared golden-vector group executed by BOTH interpreters (native golden.rs and wasm control-runtime.test.mjs), a runtime unit test covering suppress-once → held-no-refire → regrant-arms-live, and both vector interpreters extended to assert the new fields.

Verified: cargo test -p pilotage-control-web (52 pass), fmt/clippy/doc clean, wasm rebuilt, control-runtime.test.mjs / control-runtime.browser.test.mjs / control-structure / control-edges / viewer-boot.browser.test.mjs all pass.

---
- <kbd>&nbsp;2&nbsp;</kbd> #184
- <kbd>&nbsp;1&nbsp;</kbd> #183 👈

### DIFF
--- a/clients/web-control/golden-vectors.json
+++ b/clients/web-control/golden-vectors.json
@@ -660,6 +660,170 @@
           }
         }
       ]
+    },
+    {
+      "name": "a gated arm or disarm press reports suppression, never silence",
+      "steps": [
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": []
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": false
+          },
+          "expect": {
+            "motionGated": true
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": [
+              9
+            ]
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": false
+          },
+          "expect": {
+            "arm": false,
+            "armSuppressed": true,
+            "motionGated": true
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": [
+              9
+            ]
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": false
+          },
+          "expect": {
+            "arm": false,
+            "armSuppressed": false
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": []
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": false
+          },
+          "expect": {
+            "armSuppressed": false
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": [
+              9
+            ]
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": true
+          },
+          "expect": {
+            "arm": true,
+            "armSuppressed": false,
+            "motionLive": true
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": []
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": true
+          },
+          "expect": {}
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
+              0.0,
+              0.0,
+              0.0
+            ],
+            "pressed": [
+              8
+            ]
+          },
+          "session": {
+            "mode": "quad-pilot",
+            "motionGranted": false
+          },
+          "expect": {
+            "disarm": false,
+            "disarmSuppressed": true,
+            "motionGated": true
+          }
+        },
+        {
+          "pad": {
+            "axes": [
+              0.0,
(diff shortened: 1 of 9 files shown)


luofang34:
Adversarial subagent review: **no blocking findings**. Safety confirmed unweakened (arm still requires live authority on every path). Addressed from review: the readout no longer claims `streaming` during the neutral-recovery burst — it now reports `gated | recovering | streaming` three-state (recovering = frames flow but live authority unconfirmed, presses still suppressed). Noted-and-accepted: a press made entirely inside a profile-swap handover window stays undiagnosed (documented in code; distinct from the link-loss path this fixes); overlay flash may be overwritten by later status lines (log line + persistent readout are the durable record).

PR #184, opened 2026-07-22
PX4 arm/disarm enactment truth: consume COMMAND_ACK and surface FC refusals

### LINKED ISSUE #181: PX4 arm/disarm enactment is fire-and-forget: COMMAND_ACK unread, FC denials invisible
## Symptom

On the px4-gz backend, arm/disarm reality can diverge from what the viewer reports, with no operator-visible explanation:

- The viewer logs `disarm command sent` + `action result … accepted` while PX4 refuses: `WARN [commander] Disarming denied: not landed` (viewer telemetry kept showing ARMED for many seconds afterwards).
- An accepted arm can be silently undone ~10 s later: `INFO [commander] Disarmed by auto preflight disarming` — the operator sees ARMED blink back to DISARMED and reads it as "Enter doesn't arm".
- PX4 prearm/health refusals (`Preflight Fail: No connection to the GCS`, EKF, failsafe activations) never reach the viewer at all.

All of the above captured live at main `73a746cb` (`cargo xtask sim --fc px4-gz`, real Chromium driving the real viewer; PX4 lines from `target/xtask-sim/flight-controller.log`). Sequence observed: disarm at T+0 reported accepted → telemetry still ARMED at T+4 s → PX4 log shows the disarm was denied ("not landed") → later arm/disarm states flip-flopped against the command order while every action kept reporting `accepted`.

This is the mechanism behind "after disarm I cannot Enter-to-arm, gimbal still works": the gimbal manager responds regardless of arm state, velocity setpoints are dropped while disarmed, and the one signal that would explain the situation (PX4's denial) is discarded.

## Root cause

The PX4 adapter's command uplink is fire-and-forget and its only arm truth is the 1 Hz heartbeat:

- `adapters/px4/src/uplink.rs`: `send()` writes UDP and returns; `command_offboard_and_arm()` sends DO_SET_MODE + arm and transitions to `Commanded` unconditionally; `send_disarm()` likewise. No reply is ever read for these commands.
- `adapters/px4/src/adapter.rs` (`apply_control`): `ActionKind::Arm => { uplink.begin_arm(...); ActionResult::accepted(*action) }` — "accepted" means "UDP written", and per CTRL-01 that is the intended result semantics, but no *enactment* signal ever follows.
- `adapters/px4/src/adapter.rs` (`observe_arm_report`): arm state is folded from `heartbeat_armed` at 1 Hz — correct, honest, but multi-second-laggy and reason-free.
- The plumbing for the missing signal already exists and is unused by this adapter: `pilotage-mavlink` decodes `COMMAND_ACK` (`crates/pilotage-mavlink/src/codec/decoding.rs`, `FcMessage::CommandAck`) and the link state carries `last_command_ack: Option<CommandAckReport>` (`crates/pilotage-mavlink/src/link.rs`). The Aviate adapter consumes acks (`adapters/aviate/src/uplink/fc_replies.rs`); the PX4 adapter never reads them.

## Suggested fix

Keep the CTRL-01 result contract (`accepted` = command taken by the adapter) and #178's bijection untouched; add an *enactment* channel:

1. In the PX4 link apply path, consume `COMMAND_ACK` for `MAV_CMD_COMPONENT_ARM_DISARM` (and `DO_SET_MODE`), correlating on command id, and record `MAV_RESULT` plus the deny reason parameter.
2. Carry the outcome to clients on telemetry next to the existing `ArmReport` (e.g. `last_arm_command: { command, result, at }` in fc-state), so it ages honestly under the same stamp discipline instead of changing action-result latency.
3. Viewer: render a denial loudly — overlay + log (`FC refused arm: <result>`), and clear it on the next accepted ack or arm-state change.
4. Consider consuming `STATUSTEXT` for the human-readable denial ("Disarming denied: not landed", "Preflight Fail: …") — even log-only forwarding would have made this diagnosable from the viewer.
5. Document (or compensate for) PX4 auto-preflight-disarm: after an accepted arm with no takeoff inside `COM_DISARM_PRFLT`, the FC disarms itself; the fc-state transition should at minimum be logged by the viewer ("FC disarmed itself (auto preflight disarm)") rather than silently flipping the readout.

## Regression guard

- Fake-FC uplink test (the manual-clock fake-FC pattern already used in `adapters/px4/src/uplink.rs` tests): fake FC replies `COMMAND_ACK` DENIED to the arm → assert the adapter surfaces the denial report with the reason; ACCEPTED → assert the report clears.
- Host-side test: the denial report reaches the client telemetry sample (same route as `ArmReport`).
- Browser test: a denial report renders in the log/overlay.

## Acceptance criteria

- An arm or disarm the FC refuses becomes operator-visible within one telemetry period, with the FC's reason.
- A PX4 auto-preflight disarm is logged as an FC-initiated disarm, distinct from a commanded one.
- `ActionResult` semantics and timing are unchanged.

Related: #178 (adapter action-result contract), #160 (typed emergency disarm — shares the acknowledged-end-to-end principle), #151 (arm/reset latch). Companion issue: viewer-side silent arm-press suppression.

luofang34:
Fixes #181. Stacked on #183 (shares the main.js control-feedback region).

The PX4 uplink was fire-and-forget: arm/disarm went out as UDP, the adapter reported `accepted` unconditionally, and every FC refusal was invisible — live-captured during the investigation: viewer says "disarm … accepted" while PX4 logs `Disarming denied: not landed`, and `Disarmed by auto preflight disarming` silently un-arms ~10 s after an arm with no takeoff.

- Uplink: drains its command socket (PX4 acks a COMMAND_LONG to the endpoint that sent it), pairs MAV_CMD_COMPONENT_ARM_DISARM acks with the latest arm/disarm sent (the ack carries only the command id), logs refused OFFBOARD mode commands.
- Wire: `FcState.last_command_kind/result` carry the verdict on the fc-state lane; host maps it; adapter folds it beside the ArmReport.
- Viewer: each NEW refusal is logged + overlaid once; the arm readout shows `(FC refused arm: result N)` while fresh; malformed verdicts degrade to null without rejecting the report.
- ActionResult semantics (command taken by the adapter) are unchanged; aviate keeps logging acks and leaves the lane empty rather than carrying an unpaired verdict.

Guards: fake-FC ack-pairing test (refusal surfaces, fresh command clears the stale verdict), host wire-mapping assertions, JS decode/tracker/display tests.

Verified: workspace clippy -D warnings, fmt, doc gate (CI flags), targeted crate tests (385 pass), release build, wire/wire-wasm/telemetry JS suites, viewer-boot browser test (35 checks).

---
- <kbd>&nbsp;2&nbsp;</kbd> #184 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #183

### DIFF
--- a/adapters/aviate/src/adapter/sources.rs
+++ b/adapters/aviate/src/adapter/sources.rs
@@ -103,6 +103,10 @@ pub(super) fn fc_state_sample(
         .unwrap_or_default();
     Some(FcStateSample {
         arm_state: if report.armed { 2 } else { 1 },
+        // The Aviate uplink logs COMMAND_ACKs but does not yet pair them
+        // with the command they answer; the enactment lane stays empty
+        // rather than carrying an unpaired verdict.
+        last_command: None,
         stamp: MeasurementStamp {
             // Role is the discriminator; the id carries the configured
             // FC identity: (system id << 8) | component id.
--- a/adapters/px4/src/adapter.rs
+++ b/adapters/px4/src/adapter.rs
@@ -405,7 +405,12 @@ impl VehicleAdapter for Px4Adapter {
 
     fn sample_telemetry(&mut self) -> TelemetryBatch {
         self.observe_arm_report();
-        let fc_state = sampling::fc_state_sample(self.arm, self.arm_incarnation, self.started_at);
+        let fc_state = sampling::fc_state_sample(
+            self.arm,
+            self.uplink.as_ref().and_then(Px4Uplink::last_arm_ack),
+            self.arm_incarnation,
+            self.started_at,
+        );
         let gimbal_attitude = self
             .gimbal
             .is_some()
--- a/adapters/px4/src/adapter/sampling.rs
+++ b/adapters/px4/src/adapter/sampling.rs
@@ -274,6 +274,7 @@ impl super::Px4Adapter {
 /// telemetry, stamped under the FC-state source role.
 pub(super) fn fc_state_sample(
     report: Option<super::ArmReport>,
+    last_command: Option<pilotage_adapter_api::FcCommandAck>,
     incarnation: SourceIncarnation,
     started_at: std::time::Instant,
 ) -> Option<FcStateSample> {
@@ -284,6 +285,10 @@ pub(super) fn fc_state_sample(
         .unwrap_or_default();
     Some(FcStateSample {
         arm_state: if report.armed { 2 } else { 1 },
+        // The FC's verdict on the most recent commanded arm/disarm rides
+        // the same report provenance: enactment truth the operator needs
+        // when "command accepted" and "vehicle armed" disagree.
+        last_command,
         stamp: MeasurementStamp {
             role: SourceRole::FcState,
             // MAVLink frames are CRC-checked but unsigned: checksummed,
--- a/adapters/px4/src/adapter/tests.rs
+++ b/adapters/px4/src/adapter/tests.rs
@@ -4,6 +4,8 @@
 //! the wire, sampling authorization from the standard status, and the
 //! same gate discipline the Aviate adapter carries.
 
+mod uplink_ack;
+
 use std::net::UdpSocket;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
--- a/adapters/px4/src/adapter/tests/uplink_ack.rs
+++ b/adapters/px4/src/adapter/tests/uplink_ack.rs
@@ -1,0 +1,95 @@
+#![allow(clippy::expect_used, clippy::panic)]
+
+//! COMMAND_ACK enactment truth: PX4 acknowledges a COMMAND_LONG to the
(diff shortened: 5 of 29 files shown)


luofang34:
Adversarial subagent review: **no blocking findings**. MAVLink offsets/CRC verified against the decoder and spec via the round-trip test; addressing and identity filters verified; drain loop confirmed non-blocking and bounded. Addressed from review: the COMMAND_ACK correlation limit (arm/disarm share command id 400 — an in-flight ack can be attributed to a newer command under UDP reordering; heartbeat-derived arm state stays ground truth) is now stated in code at `apply_fc_reply`; the stale `decodeFcState` header comment now lists fields 3/4. Also: LINK-04 evidence re-recorded at baseline f75f99d4 (five pinned test sources changed) — fixes the evidence-trace-gate failure. Accepted latency nit: an ack drained this cycle surfaces next telemetry period, within the issue's acceptance criterion.

PR #185, opened 2026-07-22
Autolaunch lifecycle: truthful readiness, session discovery, visibility-deferred autoconnect

### LINKED ISSUE #182: Autolaunched viewer: stale-cert retry storms and hidden/unfocused autoconnect leave the tab dead until refresh
## Symptom

After `cargo xtask sim` auto-opens the viewer, the page frequently comes up without video and without live instruments; a manual refresh fixes it. Control is often also dead in the auto-opened tab (see #180).

## Scope of what was tested

A headless probe that navigated ~1.5–2 s after the launcher printed `session ready` (px4-gz backend, main `73a746cb`) had telemetry, painted PFD/HSI, and painted video by T+2 s, identical after a reload — so the *seconds-scale* backend warmup is not the dominant cause on this machine. But a faster open CAN land in a genuinely cold host (root cause 1 below), and the browser-side lifecycle defects (2–4) were all reproduced live.

## Root causes

1. **The readiness gate fires before the host can serve.** `LISTENING <port> <cert>` is printed the instant the UDP endpoint binds (`hosts/session-host/src/runtime.rs:144-148`), BEFORE `spawn_host_runtime(...)` builds the vehicle adapter and spawns the accept loop (`runtime.rs:153-161`). xtask's browser-open gates on exactly that line (`tools/xtask/src/session.rs:80-107`, `readiness.rs:149-172`), and the upstream stage gates prove liveness, not data flow (gazebo: `gz topic -l` contains the world; FC: log contains its ready marker — `tools/xtask/src/backend/aviate_gz.rs:169-196`). The auto-opened tab's `autoconnect=1` lands in that window on a bound-but-not-accepting host / cold adapter; a refresh moments later hits a warm pipeline.
2. **Stale tabs retry forever with dead certs.** Every launch mints a fresh certificate and pins it in the URL (`readiness.rs::viewer_url`). Tabs from previous sessions keep `autoconnect=1` + the old hash and auto-retry against whatever now owns port 4433. Observed live in `target/xtask-sim/session-host.log`: `TLS handshake failure … CERTIFICATE_VERIFY_FAILED` repeating every ~25 s, beginning before the new tab was even opened. `classifyConnectFailure` (`clients/web/reconnect.js`) treats a TLS handshake failure as a retryable transport error, so the storm never ends. A stale tab renders exactly the reported symptom (black video, dead instruments) — but note refreshing a *stale* tab cannot converge (its URL keeps the dead cert); the storm and dead twins are real damage, while "refresh fixed it" refers to the current tab (causes 1 and 3).
3. **The auto-opened tab can load hidden/unfocused and connect degraded.** `open` hands the URL to the default browser with no control over window, Space, or focus. Measured in an occluded window: `document.visibilityState === "hidden"`, a 33 ms `setTimeout` clamped to 576–1000 ms. Consequences: the 30 Hz control loop runs at ~1 Hz → the host silence watchdog revokes between ticks → endless revoke/regrant churn (motion generation observed climbing 9 → 111); if `document.hasFocus()` is false at the `ServerWelcome` moment, the lease probe (`clients/web/connect-authority.js`) silently skips motion authority; a background tab can be frozen before its module even connects (one live run: no host-side session for ~2 minutes after open). A refresh happens visible + focused, which is why it "fixes" everything.
4. **Autoconnect races the instrument-wasm load.** `startInstruments()` is async and not awaited (`clients/web/main.js:1871`) while autoconnect fires synchronously afterwards (`main.js:1889-1898`); panels stay blank until `instruments.mod` resolves (`main.js:1836-1842`). Self-heals in hundreds of ms, but widens the "nothing on screen" first impression.

Latent, not today's cause: `h264Registry` is null until the wasm loads (`main.js:944`), H.264 frames in that window are skipped silently and the decoder then needs the next IDR (`clients/web/video-h264.js:63-77`). Inert while sim hosts emit MJPEG only (`media.rs`, `stream_tag.rs::FOURCC_MJPEG`), but it becomes a startup race for the #172 H.264 path.

## Suggested fix (layered; each independently useful)

1. **Truthful readiness**: print `LISTENING` (or a second `SERVING` marker xtask gates on) only after `spawn_host_runtime` returns — i.e. adapter built and accept loop spawned — so the browser never opens onto a bound-but-dead port. Optionally gate on first telemetry/frame for the stages whose "up" signal precedes data flow.
2. **De-storm stale tabs**: classify a WebTransport TLS/certificate handshake failure as non-retryable in `classifyConnectFailure`; halt with a visible line: `certificate mismatch — this tab belongs to an older session` (dovetails with #146's typed-rejection classification).
3. **Session discovery instead of URL-pinned certs**: xtask writes a `clients/web`-servable `session.json` (host, port, cert hash, start stamp); the page fetches it when URL params are absent or the pinned cert fails. Old tabs then converge on retry or refresh, and the launcher URL no longer rots. The `LISTENING` parse already yields exactly these fields (`readiness.rs::parse_listening`).
4. **Defer autoconnect to visibility, and await the instrument wasm**: run the `autoconnect=1` connect once `document.visibilityState === "visible"` and after `startInstruments()` resolves; optionally defer only the *lease* until first focus with a visible "control idle — click to take control" state instead of the silent telemetry-only skip.
5. **Surface degraded modes**: hidden-throttled churn should be stated in the readout, not only visible as log spam.

## Regression guard

- Host test: no `LISTENING` line before the accept loop is live (probe: connect immediately after the line appears must succeed).
- Unit: `classifyConnectFailure` returns non-retryable for the TLS/cert failure class; reconnect controller halts and logs once.
- Browser test: boot with `autoconnect=1` under CDP-emulated hidden visibility → no connect until visible, then a normal session; panels ready at first paint after autoconnect (wasm awaited).
- xtask test: `session.json` equals the parsed LISTENING values.
- Manual acceptance: launch with two stale tabs open → no repeating TLS failures in the host log; refreshing a stale tab joins the current session.

## Acceptance criteria

- The auto-opened tab reaches video + instruments + (once focused) control with no user intervention.
- Stale tabs stop hammering the host and can converge to the current session with a plain refresh.
- No degraded state (telemetry-only, throttled churn, stale-cert halt) is silent.

Related: #146 (retry classification), #122 (media stalls), #172 (H.264 startup race is latent here), companions #180 / #181.

luofang34:
Fixes #182. Three cooperating defects left the auto-opened tab dead until a refresh:

1. **Host readiness lied by a window**: `LISTENING` printed at endpoint-bind, before `spawn_host_runtime` built the adapter and spawned the accept loop — the launcher's browser-open gate could land an autoconnecting page on a bound-but-dead host. The line now prints only after the runtime is accepting.
2. **Stale tabs stormed forever**: every run mints a fresh cert pinned in the URL; old tabs retried the dead hash indefinitely (live-observed CERTIFICATE_VERIFY_FAILED every ~25 s in the host log). xtask now serves the CURRENT session's parameters as `clients/web/session.json` (gitignored), and the viewer re-reads it after any non-config connect failure — any stale tab converges on the live session at its next attempt.
3. **Hidden-tab connect degraded silently**: an occluded tab clamps 33 ms timers to ~1000 ms (measured), starving the 30 Hz loop into watchdog revoke/regrant churn, and an unfocused lease probe skips motion authority. Autoconnect now defers to `document.visibilityState === "visible"` and awaits the instrument wasm so first paint has panels.

Guards: `session-discovery.test.mjs` (fail-closed manifest validation, change detection, visibility deferral incl. hidden→hidden transitions and listener detach), an xtask test pinning the manifest shape to the viewer validator's exact contract, and a new CI step.

Verified: fmt/clippy -D warnings, xtask + session-host tests (104 pass), release build, session-discovery/reconnect JS suites, viewer-boot browser test (35 checks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,9 @@ jobs:
       - name: viewer connection auto-recovery lifecycle (backoff/jitter, classification, control-suspended)
         run: node clients/web/reconnect.test.mjs
 
+      - name: viewer session discovery (manifest validation, stale-tab convergence, visibility-deferred autoconnect)
+        run: node clients/web/session-discovery.test.mjs
+
       - name: viewer control arm-edge detection (rising edges, hold-across-reconnect safety)
         run: node clients/web/control-edges.test.mjs
 
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ adapters/**/bridge/build/
 * 2.*
 *.bak
 *.orig
+clients/web/session.json
--- a/clients/web/main.js
+++ b/clients/web/main.js
@@ -72,6 +72,11 @@ import { readinessTransition, shouldLogReadFailure } from "./video-diagnostics.j
 import { bindVideoTargets, resolveVideoTarget } from "./video-routing.js";
 import { runIncomingStreamAcceptLoop } from "./uni-stream-accept.js";
 import { createReconnectController } from "./reconnect.js";
+import {
+  applySessionConfig,
+  validSessionConfig,
+  whenVisible,
+} from "./session-discovery.js";
 import { startDatagramControl } from "./datagram-control.js";
 import { loadControlShell } from "./control-shell.js";
 import { advanceMotionLease, applyMotionRecovery } from "./motion-lease.js";
@@ -313,6 +318,30 @@ function applyUrlParams() {
   if (params.has("cert")) els.certHash.value = params.get("cert");
 }
 
+/** Re-reads the launcher-served session manifest and updates the connect
+ *  inputs, so a stale tab (its URL pins an older session's certificate)
+ *  converges on the CURRENT session at its next attempt instead of
+ *  retrying a dead hash forever. A missing manifest (viewer served
+ *  without the launcher) is silent — there is nothing to converge to. */
+async function refreshSessionConfig() {
+  let fetched;
+  try {
+    const response = await fetch("./session.json", { cache: "no-cache" });
+    if (!response.ok) return;
+    fetched = await response.json();
+  } catch {
+    return;
+  }
+  const config = validSessionConfig(fetched);
+  if (!config) return;
+  if (applySessionConfig(els, config)) {
+    log(
+      `session discovery: connect target updated to ${config.host}:${config.port} ` +
+        `(cert ${config.certHash.slice(0, 16)}...)`,
+    );
+  }
+}
+
 /** Decodes a lowercase-hex cert hash string into a Uint8Array digest. */
 function hexToBytes(hex) {
   const clean = hex.trim().toLowerCase();
@@ -392,6 +421,13 @@ async function connect({ manual } = { manual: true }) {
       }
     },
   });
+  // A failed attempt may mean this tab belongs to an older session (its
+  // pinned certificate no longer matches the host on this port): re-read
+  // the launcher manifest so the NEXT attempt targets the live session.
+  // Config errors are excluded — they need the user, not a retry target.
+  if (!session.result.ok && session.result.failure?.phase !== "construct") {
+    void refreshSessionConfig();
+  }
   return session.result;
 }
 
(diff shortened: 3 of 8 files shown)


luofang34:
Adversarial subagent review: **no blocking findings**. Host LISTENING reorder verified strictly better (spawn failure surfaces via StageDied, not a hang); session.json is same-origin with no new attack surface; whenVisible listener lifecycle and instrumentsStarted.finally verified leak-free; viewer-boot test unaffected. Addressed from review: xtask now removes `clients/web/session.json` at teardown, so a manifest describing a dead session can never clobber hand-typed connect inputs (the stale-manifest F1 scenario). Accepted: the pre-write stale-read window self-heals on the next retry after the overwrite.

Issue #172, opened 2026-07-20
VID-05: gimbal camera as a selectable HD source; verify H264 hardware path; stream activation


luofang34:
Owner-directed frontend round (2026-07-20), video half. Follows GIM-04 (#168): the gz_x500_gimbal CGO3 camera publishes at /world/default/model/x500_0/link/camera_link/sensor/camera/image but is not bridged into the video pipeline.

1. Bridge the gimbal camera as a third video source, selectable as the main view — an HD (1280x720) feed, added world-side so the PX4 checkout stays untouched.
2. Verify the H.264 path end to end at HD, and that decode (and encode where applicable) uses hardware acceleration; document what the encoder actually is.
3. Clients activate only the streams they consume: a hidden/unselected view must not cost downlink bandwidth (background recording deferred). Minimal seam now; full negotiation stays VID-03 (#102).

Refs #162 (GIM-01 — the gimbal camera is the payload view), #102 (VID-03).

luofang34:
Pipeline findings + first slice (2026-07-20):

**Shipped in #173**: the gimbal HD (1280x720) camera as selectable source 2 (world-side rig welded to the CGO3 camera_link, gz sidecar `--gimbal-camera-topic`, client main-view entry), and `hardwareAcceleration: "prefer-hardware"` on the H.264 decoder with browser-default fallback. Verified live: LT+RStick moves the HD view, R3 recenters.

**Findings that reshape items 2-3:**
- The host encodes **MJPEG only** (software `jpeg-encoder`, quality 75, one encode shared across clients). The H.264 stack (`pilotage-protocol::h264`, wasm decode session, WebCodecs executor) is decode-side with **no producer anywhere** — every wire frame is tagged `MJPG`. "H264 hardware codec" therefore means adding a hardware **encoder** in the host media task (ADR-0016: VideoToolbox on macOS / NVENC-VAAPI on Linux) and flipping the FourCC at the frame writer. 720p software MJPEG works (demoed) but is the ceiling ADR-0016 warns about.
- **Stream activation seam** (item 3): the host currently fans out every source to every client, spawning per-(client,source) writers lazily; no announcement or subscription exists on the wire (`video_sources()` is never called in production). Minimal seam: a client→host active-source-set message on the reliable stream + gating the writer spawn in the media task, dropping writers on deselect. No RFC 6381 media-config needed.
[...]

PR #210, opened 2026-07-23
Video reconverges after stream loss instead of freezing until refresh

### LINKED ISSUE #209: VID-06: video must reconverge after a single uni-stream WebTransportError instead of freezing until refresh
## Context

Live session, 2026-07-23: one incoming video uni stream failed —

```
[00:07:50.879] uni stream read failed: WebTransportError (1 total this session)
```

— and video never recovered for the rest of the session (a page refresh, with a fresh session and re-arm, was the only way back), while datagram telemetry and instruments kept working.

The accept loop is NOT the defect: `runIncomingStreamAcceptLoop` correctly treats a single stream failure as one lost frame and keeps accepting (its collection-terminal path is separate and did not fire — the failure counter read 1). Streams kept arriving; the picture stayed frozen. That points at the resync contract above the transport: the H.264 path is strictly SPS/PPS-before-IDR with decoder generations, so after a lost frame the viewer needs a fresh config + keyframe to converge — and either the host never re-keys after a mid-stream loss, or the viewer's decoder state machine wedges awaiting a config that only ever arrived at stream start. Related: #122 (VID-04, bound frame completion / expose stalls) and the MEDIA-02 peer-local vs connection-fatal classification.

## Required behavior

- After any single-stream loss, the video pipeline CONVERGES back to live picture without operator action: the viewer resyncs on the next config+IDR (resetting its decoder generation), and the host guarantees a re-key within a bounded interval after a send-side stream error (or on a viewer request).
- The stall is VISIBLE while it lasts: the affected video slot shows a typed stalled/awaiting-keyframe state tied to the failure count, never a silent freeze.
- A collection-terminal error remains session-fatal and loudly reported, unchanged.

## Guardrails / acceptance

- Browser test: inject one failed stream drain mid-sequence; assert the next config+IDR restores decode (frames render again) and the stall state showed meanwhile.
- Host test: a send-side stream error is followed by a config+keyframe within the bound.
- Soak: N injected losses → N recoveries, failure counter matches, no refresh required.

SIM / NOT FOR FLIGHT.

luofang34:
## Context

Closes #209. Observed live 2026-07-23: one `uni stream read failed: WebTransportError (1 total this session)`, then video never recovered while instruments kept working — only a page refresh (new session, re-arm) restored it.

The investigation found the accept loop innocent (its per-stream tolerance worked; the failure counter read 1 and stayed there) and three real defects around it, one of which fully explains the symptom:

1. **The host silently evicts a client from ALL video when one writer dies.** A per-(client, source) writer task exits on a connection-fatal classification; the next fan-out `try_send` gets `Closed` and removed the whole client from the media registry — no log line, every source dark, and registration only happens at transport connect, so nothing short of a refresh restores video. A live session's control/telemetry are untouched: exactly the observed freeze.
2. **The H.264 decode session died permanently on any platform decode error.** A lost access unit leaves the next delta referencing a missing frame; WebCodecs reports a decode error; `platform_failed()` poisoned the session forever — no await-keyframe recovery existed. (Latent on the MJPEG-only host today; real for the H.264 sources of #172.)
3. **A stalled source was a silent freeze.** Nothing watched frame arrival per source, so a dead writer or dead route just left the last picture up.

## What changed

- **Host (`media.rs`): bounded writer respawn, never client eviction.** A closed sink respawns its writer — the connection may be perfectly healthy — up to `MAX_WRITER_RESPAWNS` (3), each respawn logged; beyond the bound that ONE source retires loudly (`error!`) while other sources and the client live on. Client removal belongs solely to connection teardown's deregistration, where it already happens. The MEDIA-02 per-stream classification is untouched.
- **Decode session (`pilotage-protocol::h264`): decode errors recover, capability failures stay fatal.** `platform_decode_error()` re-enters unconfigured — the generation advance retires the errored decoder's callbacks, deltas drop, and the next decodable keyframe reconfigures — bounded by strikes that only a PAINTED output frame (`note_output`, stale-generation-proof) re-arms; exhausting them fails closed. `platform_failed()` remains the permanent path for no-decoder/configure failures. The viewer executes the new actions: errored decoder closed, typed `awaiting a fresh keyframe` log, a stall banner on the slot, painting resumes on the next keyframe.
- **Frame-arrival freshness (`video-stall.js` + readout wiring):** per-source last-arrival tracking; a source silent past 2 s banners `video stalled — no frames arriving` on its slot and logs ONCE per transition; the next delivered frame repaints and logs `frames resumed` once. Driven off its own interval; reset with the session's video diagnostics.

## Guardrails

- Rust: decode-error recovery arc, strike exhaustion fails closed, painted-output re-arm, stale-generation output re-arms nothing; writer-exit respawn-then-retire policy bound.
- JS (`video-h264.test.mjs`): sync throw and async error-callback recovery arcs with positive controls (fresh keyframe reconfigures and PAINTS), stall banner painted, strike exhaustion reaches the permanent fail-visible marker, stale-callback hazards preserved.
- JS (`video-stall.test.mjs`, CI-wired): stall enters once at the threshold, never repeats, recovers exactly once, can re-stall.
- Size ratchet: new module capped at birth; `cockpit-readout.js` and `video-h264.js` caps adjusted with the change.

No evidence-graph-bound source is modified. Local gates on the branch in an isolated worktree: workspace `fmt --check`; clippy `-D warnings` (protocol, session-host, instruments-web); all Rust suites; wasm rebuild; every node suite green.

Live check worth doing at next flight: kill a viewer's writer mid-session (or watch a natural stream loss) — the slot banners, the host respawns, and the picture returns without a refresh.

SIM / NOT FOR FLIGHT.


luofang34:
Review findings verified and addressed at 2da90c6 — four of five were real; here is the disposition of each:

**P1 — recovery unbounded:** partially real, and the fixable core is fixed. The unfixable half rests on a premise the investigation overturned: the host emits MJPEG only (every frame self-contained), so there is no host-side keyframe to request or guarantee — a re-key protocol has no counterpart until an H.264 producer exists (#172 is where that lands). What IS now bounded, client-side: recovery attempts were already strike-bounded (3 without a painted frame, then permanent fail-closed), and this round makes awaiting-keyframe **stall-visible within 2 s** — the freshness watch now counts PAINTED frames only, so a source whose deltas keep arriving with nothing decodable banners and logs instead of freezing silently. `video-h264.test.mjs` pins that an errored feed reports no paint and a decoded frame reports exactly one.

**P1 — structure gate:** real. `fan_out_frame` split via `deliver_to_sink` extraction (67 lines), and the media tests moved to `media/tests.rs` to bring the file under 500. `check-structure.sh` passes locally.

**P2 — lost deregistration retains the client forever:** real. Two fixes: a dropped register/deregister command now warns instead of `.ok()`-swallowing, and fan-out gained the **all-sources-retired backstop** — once every source of a client has retired (a dead connection's writers retire within a few frames), the client is removed with a log line, restoring the old eviction guarantee with bounded patience while never touching a client with any live source.

**P2 — stall timer gated on wasm load:** real. The interval now starts at readout boot, independent of the instrument wasm — MJPEG stalls surface even when the wasm fails.

**P2 — only the pure counter tested:** real. `media/tests.rs` now drives `deliver_to_sink` with **real closed mpsc channels**: a closed sink respawns and the triggering frame demonstrably reaches the fresh writer's channel; the exit bound retires the sink and a retired sink skips; `fully_retired` proves one live source keeps the client while all-retired releases it. The full failed-stream → banner → recovery arc across a real transport remains a live-session check (noted in the PR body); the JS layers of it are covered by the paint-callback and stall-transition suites.

Isolated worktree at the new tip: fmt, structure gate, clippy `-D warnings`, 7 Rust suites, wasm rebuild, every node suite green.

luofang34:
The remaining P2 is now tracked as #211 (MEDIA-03): reaping must come from the connection's own closed signal, not command delivery or frame flow — the dormant-source hole in the all-retired backstop is exactly its Context. The H.264 re-key deferral is recorded on #172 so it survives this thread. Nothing else from this review remains unfiled.

Issue #211, opened 2026-07-23
MEDIA-03: reap disconnected media clients from the connection's own closed signal, not command delivery or frame flow


luofang34:
## Context

Follow-up from the #210 review (refs #209). The media task learns of a disconnect only through `MediaHandle::deregister`, a fire-and-forget `try_send` onto a bounded command queue. #210 made a dropped command LOUD and added the all-sources-retired backstop — but that backstop only advances under frame flow: a writer must attempt a write to exit, a sink must retire on a delivery attempt, and `fan_out_frame` only touches the sinks of sources whose frames arrive.

The residual hole: a client with any DORMANT source (a camera that stopped, a source the adapter never produces again) keeps that sink `Live` forever — no delivery attempt, no writer exit, no retirement — so `fully_retired` never fires and a disconnected client whose deregistration was dropped is retained indefinitely: a `Connection` clone, its sinks, and writer tasks parked on their channels.

Reaping must not depend on the command queue OR on frame flow.

## Required behavior

- Deregistration becomes EVENT-DRIVEN from the connection object itself: the media task tracks each registered client's own closed signal (`Connection::closed()`) — e.g. a `FuturesUnordered`/`JoinSet` of (client, closed) the media loop selects over — and removes the client the moment its connection reports closed, regardless of command delivery or frame traffic.
- The existing command path stays as the fast path (an orderly teardown deregisters promptly); the closed-signal reap is the guarantee; the all-retired backstop remains defense-in-depth for anything else.
- Reaping logs once per client with the trigger (command / closed-signal / backstop), so the three paths stay distinguishable in an incident log.

## Guardrails / acceptance

- Test: a registered client whose deregister command is dropped AND whose sources are dormant is reaped when its connection closes; its writer tasks end.
- Test: a live client with dormant sources is NOT reaped.
- Existing sink-transition and backstop tests unchanged.
- Full repository quality and evidence gates pass.

SIM / NOT FOR FLIGHT.

luofang34:
[...]

PR #214, opened 2026-07-23
Fix live media recovery after abandoned uni streams

### LINKED ISSUE #212: MEDIA-04: one failed uni-stream read starves ALL video for the session — cancel failed readers, expose open starvation, re-attach media in place
## Context

Live session 2026-07-23, running the #210 head:

```
[01:09:40.598] video source 1 stalled: no frames for 2000 ms
[01:09:40.598] video source 0 stalled: no frames for 2000 ms
[01:09:40.597] video source 2 stalled: no frames for 2000 ms
[01:09:40.130] uni stream read failed: WebTransportError (1 total this session)
```

The #210 stall watch worked exactly as designed (all three banners within 500 ms of the loss). But the failure MODE is new evidence: ONE stream read failure, then EVERY source dark simultaneously with no auto-resume, while datagram telemetry kept flowing. That is not a per-(client, source) writer death — #210's bounded respawn addresses that — it is something session-wide on the video path.

Ranked hypotheses, code-anchored:

1. **Client-side flow-control leak (primary).** No failure path ever cancels a stream reader: `readOneUniStream`'s `finally` only untracks, and the accept loop's `onStreamFailure` drops the error after counting. A stream whose drain failed (or was abandoned mid-read) keeps its reader locked and its undelivered bytes buffered — in WebTransport that holds SESSION-level flow-control credit and a concurrent-uni-stream slot until cancelled. Leak enough (or a few large frames against the data window) and the host's `open_uni`/writes starve for the entire session: all sources stall at once, datagrams unaffected. The host's 2 s deadline then "resets" a stream that may never have finished opening — whether that path returns the credit needs verification.
2. **Host-side collective writer wedge.** `open_uni`'s first await blocks (not errors) on exhausted credit; a blocked open is only escaped by the write deadline. Verify the deadline path actually recovers when EVERY writer is credit-starved, and add observability: a writer blocked past its deadline on OPEN (as opposed to write) currently logs the same generic line.
3. **No media re-attach exists.** Whatever the cause, the viewer has NO path to request fresh video on a live session — media registration happens only at transport connect. This is the viewer-side sibling of #211's finding.

## Required behavior

- Every stream-drain failure or abandonment CANCELS the reader (typed reason), returning its flow-control credit and stream slot; teardown-tracked readers already do this on session end — the per-stream failure path must too.
- Host: a writer blocked on `open_uni` past the frame deadline is distinguishable in logs from a slow write, and the deadline path demonstrably recovers stream/data credit (test at the frame_writer seam).
- Repair regardless of cause: on sustained all-sources stall, the viewer re-attaches media on the LIVE session (a typed message or a re-register seam — share machinery with #211's event-driven registration rework).
- Repro + guardrail: extend the headless-CDP E2E rig with an injected mid-read abandonment; assert credits recover and video auto-resumes without a refresh.

Refs #209, #210, #211.

SIM / NOT FOR FLIGHT.

luofang34:
Closes #212.

## Root cause

An incoming per-frame stream could be abandoned without cancelling its reader, leaving QUIC receive credit unavailable. On the host, open-deadline stalls were not distinguishable from write stalls. If every video source then stopped painting, the viewer exposed the freeze but had no way to re-register media on the still-live session.

## What changed

- Cancel failed and abandoned incoming stream readers with machine-readable reasons, including the stream accepted during session retirement; cancellation failures remain visible without masking the original read error.
- Tag host frame deadlines as `open` or `write`, and prove a timed-out open future is cancelled before the next frame proceeds.
- Add the idempotent `MediaAttachRequest` envelope arm and route it directly to media registration on the receiving live connection.
- Retry media attachment when every observed source remains stalled, and stop retrying as soon as a frame is painted.
- Add a headless Chromium recovery chain that injects mid-read abandonment and asserts stream-credit recovery, typed in-place attachment, both sources resuming, unchanged transport identity, and no page refresh.

## Validation

- `cargo fmt --all --check`
- `cargo test --all-targets`
- docs with `-D missing_docs` and `rustdoc -D broken_intra_doc_links`
- `cargo build --release`
- repository structure, requirements, standards, no-std, timing, and USB-CDC detection gates
- all 35 `clients/web/*.test.mjs` suites, including real Chromium tests
- `buf lint schemas`
- `buf breaking schemas --against origin/main`


luofang34:
Reviewed at the branch tip; independently verified in an isolated worktree: fmt, structure gate, clippy `-D warnings`, all Rust suites, wasm rebuild, and all 35 node suites including both real-Chromium tests — everything green, matching CI.

**Verdict: merge-ready.** The four #212 requirements are genuinely met:

- Reader cancellation is COMPLETE across every path — read failure (causal error preserved), mid-read abandonment, pre-read-inactive (the `trackReader` refusal now cancels with a typed reason instead of leaking), and all four teardown variants — with cancel failures surfaced rather than swallowed.
- The open-vs-write deadline split comes with the right proof: the drop-probe test pins that a timed-out OPEN future is actually dropped, releasing its stream-allowance wait — that is the credit-recovery guarantee, not just a log label.
- `MediaAttachRequest` is well-shaped: an empty body whose identity IS the receiving connection (nothing spoofable), engine-bypassed to media registration exactly like connect-time registration, with codec and wire-shape tests.
- The recovery gate only fires when EVERY painted source stalls (per-source deaths remain #210's respawn domain), debounced to the stall threshold, disarmed by the first paint.

Two non-blocking notes for follow-up:

1. **Host-side attach throttling:** `Register` is a full entry replacement (sink teardown, writer respawn on next frames). The client's 0.5 Hz cadence is polite, but the host trusts it — a per-client debounce in `media_loop` (ignore re-registration within ~1 s) would close the WEB-03-style gap cheaply.
2. **The 'never had video' boundary:** the gate needs ≥1 painted source, so a DROPPED connect-time registration never self-repairs — and cannot, since without a media capability advertisement the client can't distinguish 'registration lost' from 'telemetry-only host'. Worth a line on #211/#213 where the advertisement question already lives.

The browser chain is honestly scoped: it proves the production modules compose in real Chromium — real stream cancel with the typed reason, real wire bytes decoded to arm 19, same session token, no navigation — with QUIC's actual credit return being the browser's contract once cancel fires, which is the correct client-side obligation.

PR #216, opened 2026-07-23
MEDIA-05: reap mid-open video streams without leaking credit

### LINKED ISSUE #215: MEDIA-05: abandoning a mid-open video stream FINs a headerless stream the browser can never credit back — open starvation freezes all video until reconnect
## Context

The video freeze reproduced ON the #214 head (session-host log, 2026-07-23T01:53:50Z–01:56:37Z): connect at 01:53:51, arm at 01:54:02, and at **01:54:13 — eleven seconds after arm, right as takeoff inflates all three cameras' JPEGs — all three writers wedge within 4 ms** and stay wedged for 2.4 minutes (4296 drop-to-latest warnings, 153 deadline stalls: **148 phase="open", 5 phase="write"**) until the operator gives up and reconnects. Zero peer-stop lines. The client's re-attach cannot help; only a NEW connection recovers. This is a different defect from #212's, exposed by #214's own phase instrumentation.

## Root cause (pinned in source)

`deliver_frame` guards the whole open with one timeout: `timeout_at(deadline, channel.open())`, where `channel.open()` is TWO stages — `open_uni().await` (quinn stream-credit wait; allocation happens on resolve, so abandoning a PENDING stage 1 is clean) then `OpeningUniStream.await` (wtransport flushes the WebTransport stream-type header; blockable on the session send window).

When the deadline fires during **stage 2**, the composite future is dropped with a QUIC stream ALREADY ALLOCATED and its WT header unflushed. wtransport's `OpeningUniStream` drop releases the inner quinn `SendStream`, and quinn's `SendStream::drop` calls **`finish()` — a clean FIN, not a reset** (`quinn-0.11.11/src/send_stream.rs:354`). The peer therefore receives a stream that FINs with zero bytes or a truncated header. Chrome's WebTransport can never classify a headerless stream, so it is NEVER surfaced to JS — the #212/#214 client-side cancellation machinery is structurally unreachable — and the observed accounting is that its stream credit never returns.

The write-phase already knew this hazard — `frame_writer.rs`'s own comment says "drop ... FIN, not a reset — a stalled peer never drains it" and resets explicitly — but the open path abandons by drop because the composite `channel.open()` hides the stage boundary.

The failure loop: a transient session-window stall (takeoff frame-size spike; the 5 write-phase stalls) pushes header flushes into blocking → each 2 s deadline abandons another allocated-but-headerless stream across 3 sources → the peer's uni-stream credit (~100) exhausts → every subsequent open blocks at stage 1 forever (the 148 open-phase stalls) → permanent, connection-scoped starvation that re-attach (#214) cannot touch and only connection teardown clears. Matches every observation across all three incident sessions.

## Required behavior

- **Split the stages at the `FrameChannel` seam.** Stage 1 (credit wait, allocation-free to abandon) leaves the per-frame deadline: waiting for credit only delays frames, which drop-to-latest already absorbs; give it a separate, generous starvation bound with a loud, phase-tagged error.
- **Never abandon an allocated stream by drop.** On frame-deadline expiry during stage 2 (or any path holding an allocated stream), hand the pending open to a detached reaper that awaits it and immediately RESETs the resulting stream (absorbing an opening error). Invariant: every allocated stream ends in exactly full-frame-FIN or explicit RESET — the zero-byte/truncated-header FIN class is eliminated.
- The reaper is bounded and counted; its count is a correctness signal in the log.

## Guardrails / acceptance

- frame_writer test with a stage-2-blocking fake channel: an abandoned open must RESOLVE THROUGH THE REAPER TO A RESET — extend the #214 drop-probe to distinguish finish-from-reset, which is exactly the distinction it missed.
- Soak at the wtransport layer or E2E rig: force open abandons (tiny deadline) for minutes; assert stream credit stays healthy and video keeps flowing; the 148-open-stall signature must be unreproducible.
- The write-phase reset path and #214's client cancellation stay unchanged.

Refs #212, #214; supersedes the residual of the 2026-07-23 incident chain.

SIM / NOT FOR FLIGHT.

luofang34:
Closes #215

## Root cause

The video writer timed the two-stage `wtransport::Connection::open_uni` operation as one future. Cancelling its second await abandoned an already allocated QUIC send stream while the WebTransport stream header was still flushing. Quinn then treated the dropped stream as a graceful FIN, leaving the browser with a headerless stream it could not surface or cancel and eventually exhausting stream credit.

## Changes

- Split the frame-channel seam into an allocation-free stream-credit wait and an allocated WebTransport header flush.
- Give stream-credit starvation its own 30-second bound and phase-specific diagnostics; the per-frame deadline begins only after allocation.
- Transfer every timed-out allocated open to a detached reaper that awaits completion and explicitly resets the resulting stream.
- Bound each writer to eight pending reapers and log both active and lifetime counts.
- Preserve the existing immediate reset behavior for body-write stalls and the typed peer-local versus connection-fatal classifications.
- Split classifier tests into a dedicated module to retain the repository size guardrails.

## Guardrails

- A stage-two blocking fake proves deadline expiry does not cancellation-drop the opening future, writing continues, and the reaper resets exactly once without FIN.
- A bounded-reaper test proves the hard cap, active count, lifetime count, and reset behavior.
- A real wtransport loopback soak advertises only eight uni-stream credits, forces 32 mid-open deadlines, observes an accepted RESET_STREAM for every allocated stream, and then delivers a healthy frame with a clean FIN.

## Validation

- `cargo fmt --all --check`
- `cargo test --all-targets`
- `RUSTDOCFLAGS='-D missing_docs -D rustdoc::broken_intra_doc_links' cargo doc --no-deps`
- `cargo build --release`
- `bash scripts/check-structure.sh`
- instrument requirement, certification-claim, standards-registry, trace-inventory, bare-metal target, and timing gates

### DIFF
--- a/hosts/session-host/Cargo.toml
+++ b/hosts/session-host/Cargo.toml
@@ -33,7 +33,7 @@ jpeg-encoder = "0.7.0"
 pilotage-adapter-reference.workspace = true
 serde_json = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
-wtransport = { workspace = true, features = ["dangerous-configuration"] }
+wtransport = { workspace = true, features = ["dangerous-configuration", "quinn"] }
 
 [lints]
 workspace = true
--- a/hosts/session-host/src/runtime/media/frame_writer.rs
+++ b/hosts/session-host/src/runtime/media/frame_writer.rs
@@ -10,10 +10,15 @@ use tokio::sync::mpsc;
 use tokio::time::Instant;
 use tracing::{error, warn};
 use wtransport::error::{ConnectionError, StreamOpeningError, StreamWriteError};
+use wtransport::stream::OpeningUniStream;
 use wtransport::{Connection, SendStream, VarInt};
 
 use super::{EncodedFrame, now_ns};
 use crate::runtime::stream_tag::{FOURCC_MJPEG, VIDEO_FRAME_V2, frame_video_payload_v2};
+
+use reaper::OpenReapers;
+
+mod reaper;
 
 /// Longest a single frame's uni-stream write may take before the stream
 /// is reset and the writer moves on. A client that stops consuming one
@@ -24,6 +29,11 @@ use crate::runtime::stream_tag::{FOURCC_MJPEG, VIDEO_FRAME_V2, frame_video_paylo
 /// Generous against transient congestion (a healthy frame completes in
 /// milliseconds on the deployment link).
 const FRAME_WRITE_DEADLINE: Duration = Duration::from_secs(2);
+
+/// A peer that withholds uni-stream credit for this long is starving the
+/// connection, not merely delaying one frame. This stage allocates no stream,
+/// so cancellation is safe and later frames may retry loudly.
+const STREAM_CREDIT_STARVATION_BOUND: Duration = Duration::from_secs(30);
 
 /// Application error code carried on the RESET_STREAM of a
 /// deadline-exceeded frame. Informational to the peer, which discards the
@@ -147,9 +157,15 @@ trait FrameStream {
 /// Opens per-frame outbound streams on a connection.
 trait FrameChannel {
     /// The stream this channel opens.
-    type Stream: FrameStream;
-    /// Opens one host-initiated uni stream.
-    fn open(&self) -> impl Future<Output = Result<Self::Stream, StreamError>> + Send;
+    type Stream: FrameStream + Send + 'static;
+    /// The allocated stream while its WebTransport header is being flushed.
+    type Opening: Send + 'static;
+    /// Waits for stream credit, returning only after allocation.
+    fn request_open(&self) -> impl Future<Output = Result<Self::Opening, StreamError>> + Send;
+    /// Flushes the WebTransport stream header for an allocated stream.
+    fn finish_open(
+        opening: Self::Opening,
+    ) -> impl Future<Output = Result<Self::Stream, StreamError>> + Send + 'static;
 }
 
 impl FrameStream for SendStream {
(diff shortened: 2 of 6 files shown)


luofang34:
Reviewed at the branch tip; independently verified in an isolated worktree: fmt, structure gate, clippy `-D warnings`, all 78 host tests including the real-transport soak (70 ms) — green, matching the 4/4 CI.

**Verdict: merge-ready.** This lands the #215 design faithfully, and the parts that had to be exactly right are:

- **The stage split is at the correct boundary.** `request_open` (quinn credit wait — allocation happens on resolve, so abandonment is provably clean) sits under its own 30 s starvation bound with a distinct `open_credit` phase; the frame budget starts only after allocation. The old composite that hid the boundary is gone.
- **No allocated stream is ever abandoned by drop.** The `finish_open` future is polled by reference under the deadline and, on expiry, transferred STILL-LIVE to the reaper — which awaits it and sends an explicit RESET. The write-stall immediate-reset path and the peer-local/fatal classifications are preserved unchanged.
- **The reaper is bounded with backpressure, not just a cap:** the permit is reserved BEFORE allocation, so a wedged connection holds at most 8 reaper-owned streams and the writer waits rather than minting more — pacing open attempts into the wedge instead of amplifying it. Active + lifetime counts are logged, with an invariant check.
- **The soak test is the real proof, at the right layer.** A genuine wtransport loopback whose client advertises only 8 uni-stream credits, then 32 forced mid-open deadlines: under the old FIN-on-drop behavior credit exhausts at 8 and the test hangs, so 32 passing cycles — each observing the real QUIC stream ACCEPT its RESET — is the unreproducibility proof for the 148-stall signature. The healthy frame afterward closes the loop with a clean FIN read by the real client endpoint. And by eliminating headerless FINs at the source, Chrome's specific WT-layer accounting for them no longer matters.
- The fake-channel suite carries the finish-vs-reset distinction the #214 drop-probe missed (drop-tracking tallies for finished/reset, a stalled write resets exactly once, next frame proceeds).

One observability nit, not blocking: while all 8 reaper slots are pending, the writer parks in `reserve()` and logs the bound-reached warning once per delivery attempt — a long park is visible mainly through the fan-out's drop-to-latest warnings. If a future incident needs it, a periodic heartbeat while parked would make that state self-describing.

Recommended before or shortly after merge: one live px4-gz session with the #215 trigger (arm → take off, which is what spiked the frames) confirming the wedge no longer latches — the log signature to watch is `open_header` stalls WITH reaper resets and no `open_credit` starvation.

PR #218, opened 2026-07-23
MEDIA-06: bound per-client video bandwidth

### LINKED ISSUE #217: MEDIA-06: unpaced 3-source video overruns the receiver's UDP socket — host→client collapse at connect+21 s freezes video AND telemetry; add adaptive send budget
## Context

Video stalled again on the merged #216 head (session 2026-07-23T02:57:54Z–02:59:17Z). The #216 machinery worked exactly as designed — 54 header-flush deadlines all transferred to reapers, no credit-leak spiral, writer paced — and the telemetry it produced, plus this session's new symptoms, finally locate the true upstream cause.

**The forensic picture:**

- Wedge onset at **connect+21 s — identical to the previous session's connect+22 s — with NO arm and no takeoff this time.** The frame-size-spike trigger theory is dead; the constant is a RAMP, not a flight event.
- During the wedge: body writes block, 9-byte header flushes block and NEVER complete (29 reapers parked until teardown resolved them `not connected`), zero peer-stops — and this time **telemetry datagrams stopped reaching the viewer too** (instrument watchdog X-out) while a client→host profile activation was accepted mid-wedge at T+50 s.
- Datagrams bypass stream and connection flow control. Streams AND datagrams dying together, one direction only, with the reverse direction healthy, is not stream accounting — it is **wholesale host→client packet loss**.
- The kernel agrees: `netstat -s -p udp` on this machine shows **73,918,275 datagrams "dropped due to full socket buffers"** (cumulative across the incident sessions).

**The mechanism:** the media pipeline has NO rate control — it encodes three cameras at full rate (3 × 30 fps × quality-75 JPEG ≈ tens of Mbit/s in ~1250-byte packets, thousands of pps) and trusts QUIC. The receiver's UDP socket buffer overflows the moment the renderer's drain hiccups; loss triggers retransmission bursts that AMPLIFY the rate; the congestion equilibrium never recovers because the offered load never decreases. T+21 s is the congestion-window ramp time to the cliff. Telemetry datagrams are collateral of the same socket.

**Why now:** the freeze family began after GIM-04 (#169) added the gimbal camera — the third source, +50 % aggregate bitrate — which is what pushed loopback over the receiver's drain rate. Two-source sessions historically survived.

**The layer history, for the record:** #210 fixed writer-death eviction; #214 fixed reader-credit hygiene and added re-attach; #216 fixed the stream-credit leak death spiral. Each was real; each was downstream of this.

## Required behavior

- **A per-connection video send budget with adaptive degradation.** The host bounds aggregate encoded video bytes/sec per client and sheds load BEFORE the transport does it destructively: reduce fps and/or JPEG quality per source under pressure, driven by the loss signals that already exist (`still busy` drop counters, deadline stalls, reaper occupancy). Degradation and recovery are logged as transitions.
- Priority under pressure is explicit: telemetry datagrams and the control plane must never compete with video for the socket — video sheds first, entirely if necessary (a typed "video degraded/suspended: bandwidth" state the viewer surfaces, distinct from the stall banner).
- Host socket send-buffer sizing reviewed (quinn socket config) and documented; note that the CLIENT's receive buffer is Chrome's and not ours to fix.
- Verification: a loopback soak driving 3-source full-rate video against a slow-draining client must converge to a degraded-but-stable rate with telemetry loss-free — the T+21 s collapse signature must be unreproducible.
- The architectural fix remains #213 (WebRTC congestion control exists precisely for this); this issue is the transport-agnostic stopgap the current pipeline needs regardless.

Refs #216, #215, #212, #213; the gimbal-source timing implicates #169's added load, not its correctness.

SIM / NOT FOR FLIGHT.

luofang34:
Closes #217.

Investigation
- The incident counter identifies receiver-side UDP socket overflow: 73,918,275 datagrams dropped due to full socket buffers.
- The host encoded three 30 fps sources at fixed JPEG quality and admitted every encoded frame with no per-client aggregate limit.
- The existing capacity-one handoffs, write deadlines, and reapers bounded individual streams but supplied no feedback to the fan-out controller.
- Raising the host send buffer cannot enlarge Chrome's receive buffer and would permit larger bursts, so the host keeps the OS default.

Implementation
- Add a fair per-connection aggregate token bucket with an AIMD controller. Busy handoffs, write deadlines, live reapers, QUIC loss, and congestion events halve the budget; quiet paths probe upward.
- Shed complete video frames before QUIC, including a bounded full-suspension state at persistent minimum-rate pressure.
- Give reliable control/session streams priority +10 and video streams priority -10. Quinn schedules datagrams ahead of stream frames.
- Publish normal, degraded, and suspended bandwidth state over the reliable session-events stream and surface it separately from source stalls in the viewer.
- Keep media re-attachment on the live session while preserving the original status owner.
- Split media encoding and sink state into focused modules to retain the Rust structure bounds.

Guardrails
- Add a real loopback WebTransport soak with three 320x240 sources at 90 aggregate fps, a slow four-stream client, convergence below 500 kB/s, and 100 ordered telemetry datagrams received loss-free.
- Add controller tests for aggregate bounds, source fairness, all pressure inputs, suspension, and quiet recovery.
- Add wire round-trip, stream-priority, reaper-occupancy, viewer state, and structure-gate coverage.

Validation
- cargo fmt --all --check
- cargo clippy --all-targets -- -D warnings
- cargo test --all-targets
- cargo doc --no-deps with missing-docs and broken-link denies
- cargo build --release
- all 37 web suites, including four real Chromium suites
- structure, requirements, claims, standards, trace, no_std, target timing, wasm, buf lint, and buf breaking gates
- USB CDC scan: no connected target; conservative timing model applies

SIM / NOT FOR FLIGHT.


luofang34:
Reviewed at the branch tip; independently verified in an isolated worktree: fmt, structure gate, clippy `-D warnings`, all Rust suites (7 result blocks) including the new three-source soak (1.2 s), wasm rebuild, and all 37 web suites — green, matching the 4/4 CI.

**Verdict: merge-ready.** This is the strongest piece of the media stack to date, and it lands the #217 design faithfully:

- **The controller is the right shape and provably recovers.** AIMD with halve-on-any-pressure per 250 ms window and +250 kB/s per 5 s of quiet; the bounded suspension recovers BY CONSTRUCTION — at rate 0 nothing sends, so no pressure can accrue and the 5 s quiet path must re-open at the minimum rate. Per-source fair sub-buckets with 1 s idle expiry keep one source from starving the others; counters are wrap-aware; state publishes on transition only.
- **The feedback gap is genuinely closed:** every pressure signal the earlier fixes created (busy drops, deadline stalls, live reaper occupancy) plus quinn path loss/congestion now drives the fan-out controller — the 'bounded individual streams but no feedback' hole named in the body.
- **SESSION_EVENTS is a wire-compatible rename** of AUTHORITY_EVENTS (same 0x01 tag) broadened to carry `VideoDeliveryState` on the ordered control-priority stream — clean reuse, no collision, and the viewer surfaces it distinctly from per-source stall banners.
- **Priority split is real:** +10 control / −10 video through `SendStream::set_priority`, with datagrams scheduled ahead by quinn — and the soak PROVES it: 100 sequential telemetry datagrams received in order, loss-free, WHILE video sheds under a drain rate engineered between the minimum budget and the offered load (the fixture self-asserts both bounds, so it cannot pass vacuously). Convergence to stable Degraded ≤500 kB/s with no false suspension, bounded shutdown.

Non-blocking notes:

1. **Encode still runs for frames admission then sheds** — bandwidth is the target and it is bounded, but a suspended client still costs 90 fps of JPEG encode; a future skip-encode-when-over-budget would save the CPU too.
2. The soak proves quinn-level scheduling on loopback; the kernel UDP-buffer overrun itself is not simulated (not portably possible) — bounding the aggregate IS the defense, so the live three-camera flight remains the final acceptance: watch for `degraded` transitions in place of the T+21 s collapse, and a before/after `netstat -s -p udp` full-socket-buffer delta.
3. Recovery cadence from floor to ceiling is ~40 s (+250 kB/s per 5 s) — worth knowing when reading live logs, deliberate and fine.

With #216 (leak-free abandonment) plus this (offered-load bounding), both halves of the freeze mechanism — the death spiral and its trigger — are closed.

Issue #256, opened 2026-08-03
The closed 192-byte state ABI cannot describe sources with different group sets (ADR-0029 extensible groups)


luofang34:
The packed instrument state ABI (`crates/pilotage-instrument-state/src/abi.rs`, v5) is a closed 192-byte layout: a fixed set of groups, scalars and enums only. Every source must be described in exactly those slots.

That works while the only producer is a flight controller. It does not survive the source mix ADR-0026 already anticipates. A certified panel bridged as a `data-gateway` provides GNSS position, pressure altitude, ADS-B traffic, and a flight plan, but no airspeed, no vertical speed, and no magnetic heading; the flight controller provides attitude, rates, and airspeed but no traffic and no flight plan. Neither is a subset of the other, and "absent" has to mean honest `Missing`, never a slot that happens to read zero.

ADR-0029 already names the direction — "an extensible **state-group contract**: typed signal groups with validity and age, declared per panel, replacing the closed fixed-layout ABI" — and it is the gate in front of the rest of the multi-source work. luofang34/Pilotage#248 is the same wall hit from the display side (no string capability, so an HSI cannot show the active waypoint ident).

## What has to be decided

- How a source declares which groups it supplies, so an unfed group renders `Missing` by construction rather than by a producer remembering to flag it.
- How a panel declares which groups it *requires*, per the ADR-0029 registry, so a layout can refuse to place a panel nothing feeds.
- How groups grow without a wire break, and what a version bump costs across the wasm boundary, the FFI boundary, and the JS writer that mirrors the layout today.
- Whether variable-length content (idents, luofang34/Pilotage#248) is part of this contract or a separate capability.

## Why it is worth doing before the multi-source work

The feeder (#252), the composite adapter (#254), and any second panel type all encode assumptions about what a group is. Building them against the closed ABI means writing the assumptions twice.

## Acceptance

- A source with a genuinely different group set than the FC can be represented without adding fields no other source uses.
- An unfed required group renders `Missing` without the producer opting in.
- The no_std and no-alloc constraints of the instrument family survive the change.

luofang34:
Recording the extraction dependency, since it changes what this issue is gating.

[...]

Issue #254, opened 2026-08-03
Composite adapter: an aircraft with both a flight controller and a certified panel has no representation


luofang34:
ADR-0026 defines `full-authority` and `data-gateway` profiles, and says a data-gateway host — "a bridge to an AHRS/GNSS/ADS-B receiver, including flight-plan send/receive where the device supports it" — is an ordinary adapter family with no new protocol. Each profile alone is served today.

The combination is not. An aircraft can carry both a flight controller (Aviate/PX4) and a certified panel (a Garmin navigator over Connext) at the same time, and the session host holds exactly one adapter:

- `hosts/session-host/src/cli.rs:27` — `pub adapter: AdapterKind`
- `hosts/session-host/src/runtime.rs:99` — `fn build_engine<A: VehicleAdapter>(adapter: &A, options: RuntimeOptions)`

One host, one adapter, one adapter family. There is no way to expose an FC and a panel on the same `VehicleId`.

## Why a composite adapter rather than N adapters per host

ADR-0019 already frames the unit as "one adapter (per vehicle family) selects among link bindings". Read the *aircraft* as the vehicle family and the FC and the panel as two links to it, and the existing shape holds: a composite adapter owns both links, advertises the union of their capabilities (the FC's actuator scopes plus the panel's zero scopes), and routes control frames only to a link that advertises actuator scopes.

That keeps the host, the protocol, and the lease/fencing machinery untouched. Making the host hold N adapters instead would drag in capability aggregation and vehicle→adapter routing for no additional capability.

## The important constraint

The composite adapter **tags, it does not arbitrate**. `sample_telemetry` returns samples from both links carrying distinct `SourceId`/`SourceRole`; which source wins per function, and what happens when two sources disagree, belongs upstream in the feeder (#252) where `source_monitor`'s comparators and `MiscompareFault` already live. An adapter that quietly picked a winner would put display-critical policy in the transport layer.

## Acceptance

- A composite adapter exposes one vehicle backed by two links, advertising the union of capabilities.
- Control frames addressed to an actuator scope reach the FC link; the panel link is never asked to actuate.
- Telemetry from both links reaches the host tagged with distinct source identity, with no arbitration inside the adapter.
- With both links live and disagreeing on a compared function, a miscompare is raised rather than one source silently winning.

Issue #252, opened 2026-08-03
Feeder logic is browser-only: move telemetry ingress, derivations, and the nav bridge into a shared leaf crate


luofang34:
The feeder that turns wire telemetry into instrument state lives only in JavaScript, so it cannot run on any shell but the browser:

- `clients/web/telemetry-ingress.js` (825 lines) — ingress gating, source routing, epoch/incarnation ordering, coherence
- `clients/web/turn-derivation.js` (101 lines) — turn/heading derivation
- `clients/web/nav-display.js` (85 lines) — wire guidance (metres/radians) into the nav group (dots), including the deviation calibration

ADR-0029 already commits to moving this behind the shared core boundary, and ADR-0032 names it as the porting debt that must land *before* the iPad shell starts. This issue is that work.

## Why it is load-bearing beyond the iPad port

ADR-0026 defines three postures: `full-authority`, `data-gateway`, and host-absent `embedded`. Source selection and miscompare are most valuable in the postures with no host — a single panel with no flight controller is exactly when honest `Missing` and a flagged miscompare matter. If arbitration stays in the host (or in browser script), the embedded posture silently loses it.

So the feeder must be a **leaf crate that every shell links**, not a host service:

- host (`full-authority` / `data-gateway`) links it
- iPad client links it through the FFI
- browser drives the same crate as wasm
- a Linux native station links it directly

## Shape

A new sans-IO leaf crate depending only on `pilotage-instrument-state`, `pilotage-frames`, and `pilotage-alerts` — no `pilotage-protocol`, no prost, no transport (the constraint ADR-0017 put on the instrument family, extended to its feeder). Input is per-source group samples with identity and acquisition stamps; output is one `AircraftState` plus per-group status and the selected-source labels `source_monitor` already produces.

`pilotage-instrument-state::source_monitor` already implements the hard part (per-function comparators across `MAX_SOURCES`, `MiscompareFault` transitions). What is missing is the ingress gate and the derivations above it.

## Acceptance

- The three JS modules above are deleted or reduced to thin callers of the wasm build.
- `clients/web/telemetry-ingress.test.mjs` and `nav-display.test.mjs` still pass against the shared implementation, so behaviour is pinned across the port rather than re-baselined.
- The crate compiles for `thumbv7em-none-eabihf` alongside the rest of the instrument family (the existing no_std CI gate).

Blocked on #253 for the identity types.

luofang34:
[...]

PR #257, opened 2026-08-03
Extract source identity into a leaf crate with the ingress admission gate

### LINKED ISSUE #253: Ingress source identity is expressed only in protocol types, so the feeder cannot stay a leaf
ADR-0018's ingress model — vehicle-scoped logical source identity, opaque 128-bit incarnation, source epoch, wrapping group sequence, monotonic acquisition stamp, clock domain — is currently expressed only in `pilotage-protocol` wire terms (prost-generated types in `crates/pilotage-protocol`, consumed by `clients/web-instruments/src/decode_envelope.rs`).

The feeder crate must apply that model, but it must stay a leaf: the instrument family is defined by having no dependency on `pilotage-protocol`, prost, or any transport (ADR-0017). Today those two requirements are in direct conflict.

## What is needed

A protocol-free source-identity vocabulary in a leaf crate — source id, incarnation, epoch, wrapping sequence, acquisition stamp, clock domain — with the wire types mapping *into* it at the transport boundary rather than the feeder reaching *out* to them.

This is the same "thin bridge module owns the mapping, and it is the only place the two vocabularies meet" pattern ADR-0017 already established for instrument state.

## Why it matters beyond tidiness

The `embedded` posture (ADR-0026) has no session protocol at all — an iPad talking straight to a panel over External Accessory. There is still source identity in that posture: the panel boots, reconnects, and can be swapped. Expressing identity in protobuf terms makes it unavailable exactly where no protobuf exists.

## Acceptance

- Identity types live in a leaf crate with no prost/protocol dependency and compile for `thumbv7em-none-eabihf`.
- The wrap-safe serial arithmetic, epoch ordering, and incarnation rules from ADR-0018 are unit-tested on the leaf types.
- `decode_envelope.rs` maps wire → leaf types and holds no ordering policy of its own.

luofang34:
Closes #253. First of a sequence toward multi-source instrument feeding; the
later steps (#252 feeder crate, #254 composite adapter, #256 extensible groups)
build on this one.

## Why

ADR-0018 defines when a measurement group may replace display state. Two things
were in the way of applying those rules anywhere but the browser:

1. The identity vocabulary lived in `pilotage-adapter-api`, which depends on
   `pilotage-protocol` and `pilotage-timing`. The feeder that must apply the
   rules has to stay a leaf (ADR-0017's constraint on the instrument family).
2. The rules themselves existed only in `clients/web/telemetry-ingress.js`.

ADR-0026's host-absent `embedded` posture makes this concrete: a tablet talking
straight to a panel has no wire, but it still has source identity — the panel
boots, reconnects, and can be swapped. Identity expressed only in protobuf terms
is unavailable exactly where there is no protobuf.

## What changed

`MeasurementClock`, `SourceIncarnation`, `SourceRole`, `SourceIntegrity`, and
`MeasurementStamp` move to the new `pilotage-ingress` crate — `no_std`,
allocation-free, zero dependencies — and `pilotage-adapter-api` re-exports them,
so every existing `use pilotage_adapter_api::…` path is unchanged. This is a
move, not a copy; there is one definition.

`SourceGate` states the admission rules once:

- wrap-safe serial ordering (RFC 1982) for both epoch and sequence, so a counter
  that wraps does not make the next value look ancient;
- incarnation compared for equality only, with authorization delegated to a
  caller-supplied predicate — an aircraft profile pins a source-issued
  incarnation, a simulator profile may accept a bounded number of unseen ones,
  and that policy difference is the caller's, not the gate's;
- duplicates, reordering, acquisition-time regressions, and unexplained clock
  domain changes refused with a typed reason;
- a refusal is counted and never refreshes state age — a stale value keeping its
  true age is honest, one refreshed by a rejected frame is not;
- acceptance reports `identity_changed`, so a caller clears the other groups
  rather than letting one display generation mix values across a source reset.

## Verification

14 unit tests covering each rule, including both wrap boundaries, epoch ordering
being meaningless across incarnations, and role disambiguating colliding ids.

Gates run locally, all green: `cargo fmt --all --check`, `cargo clippy
--all-targets -- -D warnings`, `cargo test --all-targets`, `cargo doc --no-deps`
with `-D missing_docs -D rustdoc::broken_intra_doc_links`,
`scripts/check-structure.sh`, `scripts/check-instrument-requirements.sh`, and
`cargo check -p pilotage-ingress --target thumbv7em-none-eabihf`. The crate is
added to the existing no_std CI gate so the claim cannot rot.

## Not in this PR

The gate is not yet wired into any ingress path — `telemetry-ingress.js` still
owns the live rules. Replacing it is #252, which needs these types to exist
first. Doing both at once would mix a mechanical extraction with a behavioural
port in one review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,7 @@ jobs:
           cargo check
           -p pilotage-frames
           -p pilotage-geo
+          -p pilotage-ingress
           -p pilotage-alerts
           -p pilotage-instrument-state
           -p pilotage-instrument-scene
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1006,6 +1006,7 @@ dependencies = [
  "pilotage-calibration-id",
  "pilotage-camera-calibration",
  "pilotage-frames",
+ "pilotage-ingress",
  "pilotage-protocol",
  "pilotage-timing",
  "serde",
@@ -1154,6 +1155,10 @@ dependencies = [
  "pilotage-frames",
  "thiserror",
 ]
+
+[[package]]
+name = "pilotage-ingress"
+version = "0.1.0"
 
 [[package]]
 name = "pilotage-input"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "crates/pilotage-protocol",
     "crates/pilotage-authority",
     "crates/pilotage-session",
+    "crates/pilotage-ingress",
     "crates/pilotage-input",
     "crates/pilotage-timing",
     "crates/pilotage-adapter-api",
@@ -63,6 +64,7 @@ ed25519-dalek = { version = "2", default-features = false, features = ["std"] }
 pilotage-timing = { path = "crates/pilotage-timing" }
 pilotage-protocol = { path = "crates/pilotage-protocol" }
 pilotage-authority = { path = "crates/pilotage-authority" }
+pilotage-ingress = { path = "crates/pilotage-ingress" }
 pilotage-input = { path = "crates/pilotage-input" }
 pilotage-adapter-api = { path = "crates/pilotage-adapter-api" }
 pilotage-mavlink = { path = "crates/pilotage-mavlink" }
--- a/crates/pilotage-adapter-api/Cargo.toml
+++ b/crates/pilotage-adapter-api/Cargo.toml
@@ -8,6 +8,7 @@ publish = false
 workspace = true
 
 [dependencies]
+pilotage-ingress = { workspace = true }
 pilotage-protocol = { workspace = true }
 pilotage-timing = { workspace = true }
 pilotage-frames = { workspace = true }
--- a/crates/pilotage-adapter-api/src/telemetry.rs
+++ b/crates/pilotage-adapter-api/src/telemetry.rs
@@ -3,107 +3,12 @@
 use pilotage_protocol::VehicleId;
 use pilotage_timing::SimTick;
 
-/// The monotonic clock in which a measurement's acquisition timestamp is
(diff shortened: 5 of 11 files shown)


luofang34:
Both fixes applied.

**1. `Admission` is now `#[must_use]`.** You were right that it was the one value that could be silently dropped while every read-only accessor carried the attribute. The attribute demonstrably bites: adding it turned all 12 test setup lines into warnings, which under `-D warnings` is a build failure. Those lines now go through an `establish()` helper that asserts the setup stamp was actually admitted — so a setup line that stopped establishing state can no longer leave the assertion below it passing for the wrong reason.

**2. The four normative clauses are restored** in `crates/pilotage-ingress/src/source.rs`:

- `SourceRole` — "(panels and control accept only `SourceRole::OperationalEstimate`)"
- `OperationalEstimate` — "the only role eligible for"
- `PayloadDevice` — "never eligible for control validation"
- `SourceIntegrity` — "a consumer making a safety claim must require the level the claim needs"

You framed it exactly right: the FC-specific genericization was the point of the move, but the consumer obligations *are* the spec, and dropping them while claiming to make the rules available where there is no wire would have defeated the PR's own purpose.

Gates re-run locally after the amend: fmt, clippy `-D warnings`, `cargo test --all-targets`, doc with `-D missing_docs`, and the bare-metal `thumbv7em-none-eabihf` check.

On the two forward-looking notes — both recorded on #252 so they are not lost between PRs. The `sequenceGaps` / `AcquisitionRegressed` / `ClockDomainChanged` divergence in particular means #252 must be reviewed as a behavior change rather than a port, and the integrity-downgrade observation is worth deciding before a consumer relies on `identity_changed` alone.

Issue #265, opened 2026-08-06
Canonical instrument states are private and duplicated, and nothing digests what the panels draw


luofang34:
Four different things need "a known aircraft state to render", and each one
keeps its own:

- `demo_state()` — private to `crates/pilotage-instrument-raster/src/raster/tests/frame_hashes.rs`,
  the input to the `PFD_SHA256` / `HSI_SHA256` determinism baselines;
- the browser display check — states built in client script;
- the iPad prototype's preflight check — a fourth set, built downstream because
  no public one exists;
- the panel-side conformance harness, which does not exist yet but will need
  injected `Missing` / `Stale` / `Degraded` / `Failed` per group.

Nothing forces these to agree, so "fully fed" means four things. A panel that
regresses under one set can be green under another, and a standalone bench — a
shell with no host and no link at all — has nothing public to drive at all.

## Proposed shape

One public, versioned set of canonical states in the instrument family, keyed
by stable id:

```
nothing-fed        every group absent
typical            what an ordinary source supplies
fully-fed          every group fed, asymmetric on purpose
source-unusable    every group fed, source declares itself unusable
```

Asymmetry matters and should be stated where the fixtures live: a display that
mirrors an axis passes a symmetric pattern unnoticed. A panel may contribute
extreme states of its own (unusual attitude, off-scale deviation) through the
registry, so the set grows with the panels rather than in one hand-maintained
list.

Consumers become: the determinism baselines, the render digest below, the
conformance harness, and every shell's display check.

## The digest, and why it belongs with the fixtures

Once the fixtures are one set, a single value describes what a build draws:
SHA-256 over every registered panel rendered in every canonical state, with
identities hashed alongside the bytes and every field length-prefixed so two
different panel sets cannot concatenate to the same stream.

It is worth having for three reasons, and the third is the one that matters
here:

1. It moves if and only if what an operator would see moves — geometry, a
   palette constant, a glyph, a state derivation, or which panels exist.
2. Every shell computes it from the same Rust, so **all three shells reporting
   the same value is a checkable claim** that web, iPad, and a standalone bench
   are running the same instruments. Pixels cannot be compared across backends;
   scene bytes can.
3. It is the migration invariant for this whole line of work. Promoting the
   symbology layer, unifying the panel signature, and moving a panel out of the
   workspace must all leave it unchanged. That turns "the refactor was lossless"
   from a claim into a test, before the panel is somewhere else.

[...]

PR #269, opened 2026-08-06
pilotage-sha256: dependency-free SHA-256 leaf with streaming context

luofang34:
pilotage-sha256: dependency-free SHA-256 leaf with streaming context

Moves the const-evaluable SHA-256 out of pilotage-instrument-glyphs into a leaf crate. Glyphs re-exports `sha256`, so its recorded-hash pin and downstream consumers (the avionics-link prototype at its rev pin) are unchanged. One shared compression function serves the one-shot and streaming front ends, bringing every function under the 80-line cap and retiring the sha256 structure-baseline exemption. Streaming equivalence is pinned by the FIPS 180-4 vectors, exhaustive split-point/length sweeps, a mid-block Clone fork test, and a differential test against the independent sha2 implementation over lengths 0-300. The crate joins the sans-IO no_std CI gate.

The streaming context is staged deliberately ahead of its caller: the INST-01 scene digest (#265) hashes panel-by-canonical-state streams and lands with the registry work.

Refs luofang34/Indicate#36 (AIR-OUT-003 groundwork).


luofang34:
Independent review disposition (no blockers): added Debug derive and #[must_use] on the digest surfaces, debug_asserts on the two carried-block invariants, a mid-block Clone fork test, and a differential test against the independent sha2 implementation over lengths 0–300 (dev-dependency only; the thumbv7em gate builds no test targets). Corrected the over-claiming coverage comment on the million-'a' test. The glyphs re-export is retained deliberately: the downstream iPad prototype (avionics-link) consumes pilotage_instrument_glyphs::sha256 at its rev pin and migrates to pilotage-sha256 with the digest work.

PR #270, opened 2026-08-06
Symbology gets a public boundary: pilotage-instrument-symbology (#263)

luofang34:
Pure move of the five shared symbology modules (palette, status_paint, annunciation, source_label, fixed_str) out of pilotage-instrument-panels into a leaf crate panels depend on. A panel outside the panels crate can now render honest status without vendoring anything. draw_alert_stack / draw_source_label / alert_label become pub; fmt_label! is #[macro_export]; FixedStr gains Default. Alert-stack tests stay in panels (they drive the stack through draw_pfd/draw_hsi) as alert_stack_tests.rs.

Guardrail: PFD_SHA256 / HSI_SHA256 byte-identical — the move is proven lossless. Zero evidence-graph locators touched. Crate added to the thumbv7em gate.

Refs luofang34/Pilotage#263 luofang34/Indicate#36 (AIR-FLAG-001..007).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> luofang34/Pilotage#272
- <kbd>&nbsp;1&nbsp;</kbd> luofang34/Pilotage#270 👈

Issue #263, opened 2026-08-06
Instrument symbology has no public boundary: extracting one panel forks the never-skinnable attributes


luofang34:
The whole shared symbology layer of `pilotage-instrument-panels` is private:

| module | lines | what it owns |
|---|---:|---|
| `palette` | 52 | every colour, including flag and failure colours |
| `status_paint` | 73 | dashes, red X, flag boxes — how a non-`Valid` signal looks |
| `annunciation` | 138 | the annunciation stack |
| `source_label` | 43 | source identity labelling |
| `fixed_str` | 50 | number and label formatting against the fixed glyph metrics |

`lib.rs` declares all five as `mod`, not `pub mod`, and exports only `draw_hsi`,
`draw_pfd`, `PfdConfig`, `BackgroundMode`, and `VSpeeds`.

## Why this blocks a reusable panel

The HSI is 473 lines (`hsi.rs` plus `hsi/{boxes,cdi,rose}`). Moving it out of
this crate today means copying 356 lines of symbology with it, because none of
it is reachable from outside. After that copy, two panels answer "how does this
display say the value is not trustworthy?" independently, and they drift the
first time either is touched.

[ADR-0029](../blob/main/docs/adr/0029-panel-layout-look-plugins.md) names
exactly these attributes as the ones that are never skinnable: "failure pages,
flag colors, alert semantics, and required-layer isolation". A fork is a
stronger divergence than a theme, so the private-module boundary does not
protect them — it only prevents reuse while leaving duplication as the way out.

## Proposed shape

A leaf crate, `pilotage-instrument-symbology`: `no_std`, no alloc, depending
only on `pilotage-instrument-scene` and `pilotage-instrument-state`
(`SignalStatus`). Panels depend on it; it depends on no panel.

The design decision worth getting right is that "unskinnable" should be
structural rather than documented. Split the palette into

- a **safety-fixed** set — flag, caution, failure, red X — as `const`s reachable
  from no configuration path, and
- a **themable** set the shell supplies as validated data.

Then a theme is *unable* to express a changed flag colour, rather than being
asked not to. That makes the claim testable: a panel rendered under a
deliberately hostile theme must paint a `Failed` group identically to one
rendered under the default.

## Acceptance

- Both existing panels paint every non-`Valid` status through the shared crate;
  no private copy remains in either.
- A theme cannot change what `Missing`, `Stale`, `Degraded`, or `Failed` paints,
  and a test drives a hostile theme to prove it.
[...]

PR #272, opened 2026-08-06
Safety-fixed vs themable: the unskinnable set becomes structural (#263)

luofang34:
Safety-fixed vs themable: the unskinnable set becomes structural (#263)

Splits the symbology palette into a safety set (safety.rs — FAILURE_RED, CAUTION_AMBER, ANNUNCIATION_WHITE, REFERENCE_YELLOW, BAND_CAUTION; reachable from no configuration path) and a themable set supplied as validated data (theme.rs: Theme + ValidTheme + ThemeError with measured-value context). Rules: imitation distance >= 64 from all four warning hues AND the red-through-yellow segment between them (white exempt by documented design); luma contrast >= 96 for primary against panel, box, and tape grounds; full alpha everywhere except the tape ground with its own floor; distinguishable GPS/radio-nav pair. Theme::all() destructures exhaustively so a new field cannot escape validation. Every safety-semantic paint references the safety set directly — in the symbology crate and across the panels — and a structure-gate grep forbids the palette safety aliases outside symbology. Values are identical throughout, so scene bytes and the pinned frame hashes are unchanged. Hostile-theme tests drive every rule arm to rejection.

Stacked on luofang34/Pilotage#270. Refs luofang34/Pilotage#263 luofang34/Indicate#36 (AIR-FLAG-001..007).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1588,6 +1588,7 @@ dependencies = [
  "pilotage-alerts",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
+ "thiserror",
 ]
 
 [[package]]
--- a/crates/pilotage-instrument-panels/src/hsi.rs
+++ b/crates/pilotage-instrument-panels/src/hsi.rs
@@ -6,7 +6,7 @@ use pilotage_instrument_scene::{Anchor, LayerId, PaintMode, SceneError, SceneWri
 use pilotage_instrument_state::HeadingReference;
 use pilotage_instrument_state::{NavSource, PanelData, SignalStatus};
 
-use pilotage_instrument_symbology::{annunciation, palette, source_label, status_paint};
+use pilotage_instrument_symbology::{annunciation, palette, safety, source_label, status_paint};
 
 use crate::{PANEL_H, PANEL_W};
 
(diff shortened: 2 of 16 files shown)


luofang34:
Independent review disposition (4 majors, all fixed in place): the imitation rule now checks all four warning hues plus the red-through-yellow segment between them, closing the corridor (ANNUNCIATION_WHITE exempt by documented design); every themable color now carries an alpha floor (full opacity except the tape ground), so transparent primary symbology is refused; the contrast rule covers the tape ground the digits actually sit on; and the never-skinnable rewiring is completed across the panels with a structure-gate grep forbidding the palette safety aliases outside the symbology crate. Also landed: exhaustive Theme destructuring (a new field cannot escape validation), measured-value context on every ThemeError, per-ground test coverage of each validation arm, and the palette WHITE dual-role doc. Frame hashes remain pinned through the panels sweep.

luofang34:
Post-review hardening (consolidated P-stack review, findings 5–7): the primary-contrast rule now covers the sky and ground horizon halves, and measures the tape ground composited at its declared alpha over each half — a legal tape alpha can no longer launder a bright horizon past an RGB-only check. A new rule holds FAILURE_RED and CAUTION_AMBER at least SAFETY_CONTRAST_LUMA_MIN (64) luma from the panel/box grounds they annunciate over, refusing the equiluminant red-on-green burial case. SAFETY_HUES is kept honest by a screened-or-exempt test plus a structure-gate pin on the safety-constant count, and the alias grep documents that it is a text ratchet, not a proof.

Issue #248, opened 2026-07-29
TO-waypoint ident display needs a string capability in the packed instrument ABI


luofang34:
ADR-0031 puts the active waypoint ident on the wire (`NavGuidanceState.to_ident`/`from_ident`), but the HSI cannot display it: the 192-byte packed state ABI (crates/pilotage-instrument-state/src/abi.rs, v5) carries scalars and enums only — no string capability — and `NavData` has no ident field.

Displaying the ident (standard on any HSI/moving-map: the active fix name next to DIST) needs: an ABI extension for bounded strings (fixed-length slot + validity, version bump, JS writer mirror), a `NavData` ident field with vocabulary/glyph coverage rules (the glyph pack must cover the charset; uncovered chars must fail closed), and panel layout for the readout. Fits the ADR-0029 extensible-group direction; the wire side is already done.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #273, opened 2026-08-06
Wire flip: the state boundary speaks ABI v6, v5 is deleted (#256)

luofang34:
abi_version()=6; state buffer becomes a CAPACITY-sized self-delimiting frame behind state_ptr()/state_capacity(); malformed-frame decode errors map to new append-only RenderStatus::StateMalformed=11 (mirrored in instrument-health.js). v5 (192-byte closed layout) is deleted; its two structure-baseline exemptions retire. clients/web/state-abi.js is the one JS writer (absent key = absent tag) and instruments.js delegates writeState to it, dropping the hand-mirrored offset table. The new cross-language golden suite (state-abi.test.mjs) pins the writer byte-for-byte against the same committed fixture frames the Rust codec pins (all three ADR-0026 postures + INVALID-ident marker), and joins CI. The fail-visible suite re-derives its VAL-01 assertions by walking the frame. All node suites pass against the rebuilt wasm.

Downstream: avionics-link inherits v6 at its next rev-pin advance; until then its frames fail closed as BadVersion.

Stacked on luofang34/Pilotage#271. Refs luofang34/Pilotage#256 luofang34/Pilotage#248 luofang34/Indicate#36 (AIR-IN-008/009, AIR-BAS-007).

---
This is **part 2 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;6&nbsp;</kbd> luofang34/Pilotage#277
- <kbd>&nbsp;5&nbsp;</kbd> luofang34/Pilotage#276
- <kbd>&nbsp;4&nbsp;</kbd> luofang34/Pilotage#275
- <kbd>&nbsp;3&nbsp;</kbd> luofang34/Pilotage#274
- <kbd>&nbsp;2&nbsp;</kbd> luofang34/Pilotage#273 👈 
- <kbd>&nbsp;1&nbsp;</kbd> luofang34/Pilotage#271


luofang34:
Independent review disposition (2 majors, both fixed): the fail-safe enum decode table is re-pinned for v6 — out-of-range NavSource/NavFromTo/EstimateQuality/SnapshotCoherence bytes decode to Unknown and Unknown encodes as 255, with a mutation of either arm now failing tests; RenderStatus::StateMalformed is exercised on both sides (a descending-tag frame drives render_into to the status in Rust, and the JS suite pins REASON.STATE_MALFORMED = 11 and its surfacing through renderPanel). Minors: the writer now zeroes the group count before writing so a partial write degrades to the empty frame instead of splicing headers; the capacity floor is the measured maximum frame (probe-encoded, not hand-mirrored); the v5-referencing doc clauses now state the codings directly.

PR #283, opened 2026-08-06
Panel registry: descriptors as composed data, config as bounded TLV (#264)

### LINKED ISSUE #264: Panels have no common signature or descriptor, and the critical-layer mask is already written three times
Panels are enumerated by hand in every shell, and the two panels do not share a
signature:

```rust
pub fn draw_hsi(data: &PanelData, alerts: Option<&AlertOutput>, scene: &mut SceneWriter) -> …
pub fn draw_pfd(data: &PanelData, cfg: &PfdConfig, alerts: Option<&AlertOutput>, scene: &mut SceneWriter) -> …
```

There is no panel id type, no descriptor, and no registry.

## The duplication is not hypothetical — it is already three copies

The critical-layer mask, which is a property of a panel, is written out
per-shell:

- `clients/web-instruments/src/exports.rs` — `PFD_CRITICAL_LAYERS`,
  `HSI_CRITICAL_LAYERS`, and `validate_panel_scene` dispatching on
  `match panel_idx { 0 => …, 1 => … }`;
- `clients/web/instruments.js` — `export const PANEL = { PFD: 0, HSI: 1 }`;
- a third copy now exists downstream in the iPad prototype, which had to
  reconstruct the same masks to gate a frame before visible commit.

A fourth arrives with any standalone bench. This is the four-layer enumeration
[ADR-0029](../blob/main/docs/adr/0029-panel-layout-look-plugins.md) says the
registry replaces, and the mask is the worst thing to duplicate: a shell whose
copy is stale accepts a frame that omits a band the panel considers critical,
and it looks correct.

Indices make it worse. `PANEL_PFD = 0` is positional, so it breaks as soon as a
build omits a panel or a plugin adds one — which is the whole point of the
work.

## Proposed shape

A `Panel` trait plus a descriptor, both `no_std` and alloc-free, with a
`&'static` registry the shells read:

```
id             stable string, never an index
title          for a picker or a report
required_layers  the mask, owned here and nowhere else
required_groups  blocked on #256 — a panel must be able to state what it needs
design_frame     per panel, not a crate constant
config           what data a shell may supply, and its schema
```

Two calls worth making deliberately:

- **`design_frame` belongs to the panel.** `PANEL_W`/`PANEL_H` are crate
  constants shared by both panels today. A moving map or an engine page will
  not be 480x360, and a backend that maps one fixed frame clips silently.
- **Config is validated data, not a Rust type per panel.** `PfdConfig` is
  today's exception; ADR-0029 already commits to configuration being
  declarative data consumed by web and native shells alike. Whatever shape is
  chosen has to be expressible over the wasm boundary and the FFI boundary
  without a per-panel binding.

## Acceptance

- No shell contains a panel list, a panel index, or a critical-layer mask. All
  three read the registry: the wasm boundary, the FFI boundary, and a
  standalone bench.
- Adding a panel touches the registry and nothing else.
- `validate_panel_scene` takes its mask from the descriptor.

Blocked on #256 for `required_groups`, and #248 if variable-length content is
part of the same contract. The descriptor should not freeze before those land,
because after extraction it is a cross-repo API.

luofang34:
A panel becomes a plugin over the stable contracts (ADR-0029, ADR-0033): pilotage-instrument-registry holds the PanelDescriptor a shell consumes — identity, title, required layers, required state groups (GroupSet over the wire tags), design frame, background capability, config schema, honest-status group regions, extreme-state hooks, raster-baseline slot, and the draw entry point. Registry::new validates the composition at init with typed context errors; there is no link-time registration magic.

Configuration crosses every boundary as one bounded key-TLV blob (ascending u16 keys, 256-byte cap, unknown key rejected rather than skipped). The PFD decodes it via PfdConfig::from_config; BackgroundMode::Svs { viewport, quality } is accept-and-cede — validated and carried, drawn exactly as None, pinned by byte-equivalence tests — so the SVS renderer consumes a ceded band later without any panel change.

panels/src/descriptors.rs owns the PFD/HSI descriptors and BUILTIN_PANELS; the shell critical-layer masks now read from the descriptors (the shell holds no mask of its own), pinned equal by the descriptor-vs-direct draw equality tests. ADR-0029 moves to Accepted; ADR-0033 records the registry-as-data, config-TLV, theme-floor, and digest decisions. The registry crate joins the thumbv7em no_std gate. Frame hashes unchanged.

Refs luofang34/Pilotage#264 luofang34/Indicate#36 (AIR-OUT-003, AIR-OUT-005).

---
This is **part 1 of 8 in a stack** made with GitButler:
- <kbd>&nbsp;1&nbsp;</kbd> luofang34/Pilotage#283 👈 

Closes luofang34/Pilotage#264.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,7 @@ jobs:
           -p pilotage-instrument-panels
           -p pilotage-instrument-raster
           -p pilotage-instrument-feeder
+          -p pilotage-instrument-registry
           --target thumbv7em-none-eabihf
 
       - name: target timing gate (REN-04)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1554,6 +1554,7 @@ dependencies = [
  "pilotage-alerts",
  "pilotage-frames",
  "pilotage-instrument-glyphs",
+ "pilotage-instrument-registry",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "pilotage-instrument-symbology",
@@ -1569,6 +1570,16 @@ dependencies = [
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "sha2 0.10.9",
+ "thiserror",
+]
+
+[[package]]
+name = "pilotage-instrument-registry"
+version = "0.1.0"
+dependencies = [
+ "pilotage-alerts",
+ "pilotage-instrument-scene",
+ "pilotage-instrument-state",
  "thiserror",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ members = [
     "crates/pilotage-instrument-panels",
     "crates/pilotage-instrument-raster",
     "crates/pilotage-instrument-feeder",
+    "crates/pilotage-instrument-registry",
     "crates/pilotage-evidence",
     "crates/pilotage-mission",
     "adapters/reference-headless",
@@ -99,6 +100,7 @@ pilotage-instrument-symbology = { path = "crates/pilotage-instrument-symbology" 
 pilotage-instrument-panels = { path = "crates/pilotage-instrument-panels" }
 pilotage-instrument-raster = { path = "crates/pilotage-instrument-raster" }
 pilotage-instrument-feeder = { path = "crates/pilotage-instrument-feeder" }
+pilotage-instrument-registry = { path = "crates/pilotage-instrument-registry" }
 pilotage-adapter-reference = { path = "adapters/reference-headless" }
 pilotage-adapter-gazebo = { path = "adapters/gazebo" }
 pilotage-adapter-aviate = { path = "adapters/aviate" }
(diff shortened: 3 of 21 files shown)


luofang34:
Review dispositions (consolidated registry-stack review, part 1): F1 — every present key is now validated regardless of mode, and a well-formed SVS key under a non-SVS background is refused as the new ConfigError::InertKey rather than silently dropped, with tests for malformed-inert, well-formed-inert, and out-of-frame payloads; F2 — the SVS viewport must now sit inside the design frame; F3 — the v-speed ladder requires strict ordering on the band-defining pairs (a collapsed ladder is refused, with a test); F4 — malformed extreme-state ids get their own BadExtremeId{index, position}, and Truncated documents key=0 as the before-the-key case; F5 — require_schema now fails closed on an unreadable header; F6 — DEFINED_LAYER_BITS computes through u16 with a compile-time assert. The staging note (no shell consumes ConfigBlob at this commit) is accurate — the consumption lands two PRs up-stack in luofang34/Pilotage#287.

PR #284, opened 2026-08-06
Canonical states and the cross-shell scene digest (#265)

### LINKED ISSUE #265: Canonical instrument states are private and duplicated, and nothing digests what the panels draw

luofang34:
The shared corpus lands in the registry crate: nothing-fed, typical (the raster demo fixture verbatim — the raster crate now delegates to it, and the unchanged pinned frame hashes prove the copy is lossless), fully-fed (every group at once, asymmetric values, live monitor channel, both nav idents), and source-unusable. Panels contribute stress fixtures through their descriptors.

scene_digest streams SHA-256 over a domain separator, the scene and state contract versions, and per panel and corpus state the length-prefixed panel id, state id, and emitted scene bytes — drawn with the empty config and no alerts, so the digest is invariant to SVS and themes by construction. From here it is THE cross-shell migration invariant: shells report the same digest or they are not showing the same instruments. The builtin digest is pinned in the panels crate (a registry dev-dependency back onto panels would duplicate the crate in the test graph and split its types) beside moves-on-change tests; pixel hashes stay per-backend regression tests.

Refs luofang34/Pilotage#265 luofang34/Indicate#36 (AIR-OUT-003).

---
This is **part 2 of 8 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> luofang34/Pilotage#284 👈 

Closes luofang34/Pilotage#265.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,7 @@ jobs:
           -p pilotage-instrument-panels
           -p pilotage-instrument-raster
           -p pilotage-instrument-feeder
+          -p pilotage-instrument-registry
           --target thumbv7em-none-eabihf
 
       - name: target timing gate (REN-04)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1554,6 +1554,7 @@ dependencies = [
  "pilotage-alerts",
  "pilotage-frames",
  "pilotage-instrument-glyphs",
+ "pilotage-instrument-registry",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "pilotage-instrument-symbology",
@@ -1569,6 +1570,16 @@ dependencies = [
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "sha2 0.10.9",
+ "thiserror",
+]
+
+[[package]]
+name = "pilotage-instrument-registry"
+version = "0.1.0"
+dependencies = [
+ "pilotage-alerts",
+ "pilotage-instrument-scene",
+ "pilotage-instrument-state",
  "thiserror",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,6 +1567,7 @@ dependencies = [
  "libm",
  "pilotage-instrument-glyphs",
  "pilotage-instrument-panels",
+ "pilotage-instrument-registry",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "sha2 0.10.9",
@@ -1580,6 +1581,7 @@ dependencies = [
  "pilotage-alerts",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
+ "pilotage-sha256",
  "thiserror",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ members = [
     "crates/pilotage-instrument-panels",
     "crates/pilotage-instrument-raster",
     "crates/pilotage-instrument-feeder",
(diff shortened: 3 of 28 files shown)


luofang34:
Review dispositions: F8 (blocker) was already fixed in the pushed head the CI confirms CLEAN — typical() is split into helpers and the structure gate passes at every stack commit; F9/F11 — the digest now binds the contract-relevant descriptor fields (required layers, groups, design frame, background capability, config schema) and every item carries a role tag, with a moves-on-mask-change test; F12 — an overrun writer report now fails closed as Scratch instead of digesting a truncated scene; F13 — the undersized-scratch test pins the exact error per size; F14 — GroupId::ALL replaces the hand-list; F10 — the ADR digest paragraph now names extreme states and states the real basis of theme independence. The pinned digest moved once more with this re-framing, recorded in the luofang34/Pilotage#289 re-pin note.

PR #286, opened 2026-08-06
Admission harness: foreign panels admitted by conformance, not trust (#267)

### LINKED ISSUE #267: No panel-side conformance harness: ADR-0029's admission gate does not exist
[ADR-0029](../blob/main/docs/adr/0029-panel-layout-look-plugins.md) admits
foreign panels "by conformance, not trust", and describes the harness that does
it: validate a plugin panel's scenes for the layer contract, budgets, glyph
vocabulary coverage, and honest-status rendering under injected
`Missing`/`Stale`/`Failed` inputs — "the inverse of the existing backend
conformance".

It does not exist. `pilotage-conformance` covers session and wire and has no
instrument coverage.

## Why this is blocking rather than a follow-up

The existing corpus checks the *backends*: given bytes, does an interpreter
agree. Nothing checks the *producers*: given a state, does a panel emit a scene
that is well-formed, within budget, drawn from the covered glyph vocabulary, and
honest about what it does not know.

While every panel is in this workspace and written by the same hand, that gap is
survivable. The moment a panel can be reused — the point of this line of work —
it is the only thing standing between a layout and a panel that paints a
plausible number for a group nothing feeds. A panel is admitted by this harness
or it is admitted by trust; there is no third option, and trust is what
ADR-0029 rejects.

## Proposed shape

Driven by the registry, not by a list of panels, so it covers every panel
including ones that arrive later. For each registered panel, for each canonical
state and each injected degradation:

- **Layer contract** — `validate_layers` accepts, and the scene carries every
  band the descriptor declares critical. The second half is what a shell
  currently has to re-derive; see the mask duplication in the descriptor issue.
- **Budgets** — per-layer command count, stack depth, scene bytes, text length.
- **Glyph coverage** — every text run is drawn from the covered vocabulary. A
  panel that paints an uncovered character fails the whole frame on a
  fail-closed backend, and it only shows up on the state that happens to paint
  it. This is a cheap check that currently exists nowhere: the corpus does not
  pin text rendering (`canvasMethods` is null for the text entries, noted in
  #255).
- **Honest status** — with a required group withheld, the panel must not paint
  a value for it. This is the hard one to state mechanically and the reason the
  harness is worth building carefully rather than quickly: "did not paint a
  number" has to be checkable from the scene, which likely means the harness
  asserts on decoded text runs in the region the group owns.

## Acceptance

- Both existing panels pass, driven from the registry.
- A deliberately dishonest panel — one that paints a value for a withheld
  group — fails the harness. Without that test the harness is decoration.
- The harness runs in CI as a gate, and its verdict is what "this panel may
  join an operational layout" means.

luofang34:
pilotage-instrument-conformance drives every registered panel through the shared corpus plus its own extreme states, and through every single-group withholding its descriptor declares, checking four families: the layer contract (well-formed scenes, required bands present under every degradation), budgets (a text run anchored outside the design frame without a bounding clip is a counted warning — the pitch ladder currently accounts for all 144, the frame-contract debt), glyph coverage (every text run inside the controlled vocabulary), and honest status (a withheld group's declared region may not gain visible numeric text the nothing-fed furniture lacks, and the furniture itself must be numeral-free inside every declared region so an always-fabricating panel cannot launder its numbers into the baseline). Text runs carry clip context — anchor-inside-clip is the visibility test — so tape labels scrolled past their strip's clip cannot false-flag a readout region.

The builtin descriptors gain their honest-status regions: narrow value-readout boxes keyed by the group whose data the number comes from (the altitude readout is a kinematics region — the altitude group only qualifies the datum). A cfg(test) dishonest fixture that fabricates a readout MUST fail admission, and does; a hollow panel missing its required band fails the layer family; the shipped panels pass all 64 cases.

Refs luofang34/Pilotage#267 luofang34/Indicate#36 (AIR-FLAG-001..007, AIR-IN-009).

---
This is **part 4 of 8 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> luofang34/Pilotage#286 👈 

Closes luofang34/Pilotage#267.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,7 @@ jobs:
           -p pilotage-instrument-panels
           -p pilotage-instrument-raster
           -p pilotage-instrument-feeder
+          -p pilotage-instrument-registry
           --target thumbv7em-none-eabihf
 
       - name: target timing gate (REN-04)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1532,6 +1532,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "pilotage-instrument-conformance"
+version = "0.1.0"
+dependencies = [
+ "pilotage-alerts",
+ "pilotage-instrument-glyphs",
+ "pilotage-instrument-panels",
+ "pilotage-instrument-registry",
+ "pilotage-instrument-scene",
+ "pilotage-instrument-state",
+ "thiserror",
+]
+
+[[package]]
 name = "pilotage-instrument-feeder"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1554,6 +1554,7 @@ dependencies = [
  "pilotage-alerts",
  "pilotage-frames",
  "pilotage-instrument-glyphs",
+ "pilotage-instrument-registry",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "pilotage-instrument-symbology",
@@ -1569,6 +1570,16 @@ dependencies = [
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "sha2 0.10.9",
+ "thiserror",
+]
+
+[[package]]
+name = "pilotage-instrument-registry"
+version = "0.1.0"
+dependencies = [
+ "pilotage-alerts",
+ "pilotage-instrument-scene",
+ "pilotage-instrument-state",
  "thiserror",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,6 +1567,7 @@ dependencies = [
  "libm",
  "pilotage-instrument-glyphs",
  "pilotage-instrument-panels",
+ "pilotage-instrument-registry",
  "pilotage-instrument-scene",
  "pilotage-instrument-state",
  "sha2 0.10.9",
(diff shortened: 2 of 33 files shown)


luofang34:
Review dispositions: F18 — the harness now maps every text run and clip through the scene's affine state (same post-multiplied translate/rotate math as the reference rasterizer) into design-frame space; the translate-evasion probe is a pinned regression test and fails admission. F19 — runs are conservative nominal-ink rectangles positioned by their anchor, and region/clip/frame tests use rectangle overlap, not anchor points. The re-derived ratchet is 33, and every one is real: the PFD groundspeed/baro readouts paint size-16 text whose ~107-unit nominal ink overhangs their 90-unit boxes and the frame edge (status_paint::readout_box has no fit shrink) — honestly attributed in the test comment; fixing the paint moves frame hashes and is deliberately its own change. F20 — the HSI digital heading readout and the PFD VSI numeral strip now carry declared regions; F21 — the dead furniture exemption is gone (the furniture pre-check owns that property); F22 — furniture laundering has its own DishonestFurniture variant; F23 — the furniture baseline comes from CANONICAL_STATES[0]; F24 — GroupId::ALL; F25 — the crate doc states the cooperative-not-adversarial scope, the digit-only numeral test, and the nominal-ink approximation. A new leaking-fabricator fixture pins DishonestNumeral on the fed-vs-withheld path.

Issue #255, opened 2026-08-03
Pilotage consumer compatibility gate: verify the shipped Indicate backend against the compatibility tuple


luofang34:
This issue is recast. The earlier form asked where a native CoreGraphics backend is *registered* so a corpus bump reaches it, implying a global reverse-dependency registry that Indicate would maintain for its consumers. That framing is rejected: Pilotage verifies the backend that it ships, and no repository keeps a registry of another repository's consumers.

## Decision context

- ARCH-INST-01 / ADR-0032 names **Indicate** as the contract and panel owner and **IndicateAppleDisplay** as the Apple display backend (the Swift CoreGraphics scene interpreter described in the original discussion below).
- ADR-0032 defines the compatibility tuple: state ABI; scene format; corpus version and digest; registry scene digest; screen-composition digest.

## What this issue asks now

Pilotage adds a **consumer compatibility gate**:

1. Pilotage pins the Indicate revision it ships (ADR-0034) and pins the five tuple values computed from that revision (state ABI version, scene format version, corpus version + `corpusSha256`, registry scene digest, screen-composition digest).
2. A Pilotage gate recomputes or re-reads each value from the pinned Indicate revision and fails the build on any disagreement — the existing `corpusSha256` pin pattern in `clients/web/scene-conformance.test.mjs`, extended to the full tuple. The Indicate `release-manifest.json` at the pinned rev is the comparison source.
3. Compatibility is checked **before paint**: a backend that does not match the pinned tuple does not render instruments; the failure is shown, not painted over.
4. When Pilotage ships IndicateAppleDisplay, the same tuple check applies to the Swift backend: it verifies the corpus version and digest it was built against at startup, and refuses to paint on mismatch.

Explicitly out of scope: a reverse-dependency registry in Indicate listing Pilotage (or any consumer). Sync flows through the pin and the manifest, never through upstream bookkeeping.

## Original discussion (superseded, kept for record)

The original body described the Swift CoreGraphics interpreter, its corpus conformance results (schemaVersion 2, corpusVersion 3 era), and three questions: where a native backend is registered, where the Swift package lives, and the unpinned text-rendering behavior. Outcomes:

1. *Registration:* nowhere global. Each consumer pins and verifies the tuple of the backend it ships (this issue).
[...]

PR #287, opened 2026-08-06
The shell consumes the registry: enumeration, config, and no panel list of its own (#268)

### LINKED ISSUE #279: Surface v6 DecodeReport forward-compat counters (unknown_groups / extended_groups) at the shell boundary
From the independent D-stack review of luofang34/Pilotage#273 (finding 5, dispositioned as a follow-up): `abi/v6.rs` computes `DecodeReport.unknown_groups` / `extended_groups` and tests pin both, but `clients/web-instruments/src/exports.rs` drops them at both decode sites (`Ok(report) => report.state`). A producer emitting tags this build cannot place is invisible in the field — asymmetric with the scene path, which surfaces exactly this as `InstrumentModule.unknownOpcodes`. The v6 doc calls an unknown tag "a counted skip"; today it is only a skip.

Natural landing spot: the INST-01 shell-rewiring PR (#268 P7), which already extends the exports enumeration/diagnostics surface — mirror the unknownOpcodes pattern there and read it into the instruments.js health object.

Refs luofang34/Indicate#36 luofang34/Pilotage#273.

luofang34:
The wasm surface gains the registry enumeration (panel_count/id/title/required_layers/required_groups/design_width/design_height/background_capability) and set_panel_config (framing- and schema-validated; a refused blob leaves the previous configuration in effect), all append-only. Rendering goes through the descriptor draw entry: the panel index resolves against the composed registry, the per-panel config blob rides to the panel, and a config refusal surfaces as the new RenderStatus::ConfigInvalid = 12 mirrored in instrument-health.js. set_v_speeds stays as sugar that splices the V_SPEEDS entry into the PFD blob. The v6 DecodeReport forward-compatibility counters are surfaced (state_unknown_groups / state_extended_groups), closing luofang34/Pilotage#279 beside the scene path's unknownOpcodes.

instruments.js derives its panel map and per-panel design frames from the enumeration at load — the hard-coded PANEL and LOGICAL_W/H are gone, and a pin test keeps the derived map equal to {PFD: 0, HSI: 1} so consumers cannot drift. Missing enumeration or descriptor nonsense fails load as ABI_MISMATCH, fail-visible as ever.

Closes luofang34/Pilotage#279. Refs luofang34/Indicate#36 luofang34/Pilotage#255 (AIR-OUT-003, AIR-OUT-005).

---
This is **part 5 of 8 in a stack** made with GitButler:
- <kbd>&nbsp;5&nbsp;</kbd> luofang34/Pilotage#287 👈 


luofang34:
Review dispositions (shell-stack review): F1 (blocker) — the cockpit and the demo page now key their health tables and targets by the page's own canvas ids (equal to the descriptor ids by contract), which exist before any wasm loads, so the failure paths paint visible covers pre-load and on load failure; the module resolves the id dialect internally in renderPanel, and the composition rides the module instance (mod.panels). The instrument-health set helpers fail visibly (orphanTarget cover) on a missing entry instead of throwing, with pre-load construction tests pinning cover/tick/fail. F2 — enumeration builds into locals and swaps only after full validation. F3 — stateUnknownGroups()/stateExtendedGroups() are reachable on InstrumentModule with a test; luofang34/Pilotage#279 is genuinely closed now. F4 — splice_v_speeds iterates the panel's own schema (Registry-validated ascending), with preserve/replace/clear/malformed unit tests. F5 — set_v_speeds resolves the PFD by descriptor id and gates the spliced blob through require_schema. F7 — the forward-compat counters count once per frame step, not per panel render. F8 — the render path borrows the config slice instead of cloning per frame.

PR #300, opened 2026-08-06
Counters proxy: reads, spread, and Object.keys agree (#299)

### LINKED ISSUE #299: counters proxy: an unknown assigned field is readable but invisible to spread and Object.keys
Minor, filed for completeness rather than urgency.

`AvionicsIngress#counters` (`clients/web/telemetry-ingress.js:269-293`) is a
Proxy standing in for what used to be a plain object. #290 added `ownKeys` and
`getOwnPropertyDescriptor` so spread and enumeration see the counter fields
again. That works — probed against the shipped module:

```
spread keys: 15
after patch, direct: 7 in spread: 7
```

Patched values are visible to spread, and `lastRejectReason` is correctly
absent from the enumeration (it was never a counter) while remaining readable
directly.

## The asymmetry

`ownKeys` enumerates `Object.keys(this.#rawCounters())` — the fifteen known
counters. The `set` trap accepts *any* field name into `#counterPatches`, and
`diagnostics()` then writes it onto the returned object, so `get` finds it.
An unknown field is therefore readable but invisible:

```
bogus field direct: 3   in spread: undefined   in keys: false
```

Only reachable by assigning a field the counters object never had, so nothing
in the shell hits it. It is a small inconsistency in an object whose whole
purpose is to behave like the plain object it replaced.

## Ask

Key `ownKeys` off the same set `get` resolves against — the patched
`diagnostics()` field set minus `lastRejectReason` — so read, spread, and
`Object.keys` agree. Alternatively reject a `set` to an unknown counter, which
is arguably the more honest of the two: a write to a counter that does not
exist is a caller bug, and silently accepting it is how it stays one.

Refs #252 #290.

luofang34:
ownKeys resolves the same set get does — the counter fields plus any patched-in assignment — so a field a caller assigns is enumerable exactly as it would be on the plain object this proxy stands in for; lastRejectReason stays readable but out of enumeration, as it was never a counter. Pinned by the spread test.

Closes #299.

---
- <kbd>&nbsp;2&nbsp;</kbd> #301
- <kbd>&nbsp;1&nbsp;</kbd> #300 👈


luofang34:
Independent review disposition: the trap set now resolves one own-property model — a `has` trap (so `in` agrees with reads, spread, and Object.keys), `getOwnPropertyDescriptor` answers undefined for nonexistent keys instead of fabricating descriptors, and lastRejectReason is a listed own property held out of enumeration (enumerable: false), which every operator now agrees on. Pinned by a membership test covering in/hasOwn/descriptors/keys. The turn-derivation cap line moved to #301 (this PR now carries 90, its own tree's size). The bound LINK-AUTHZ evidence result is re-recorded over the real 37-check run with the config anchor on this branch's lane commit.

PR #301, opened 2026-08-06
Two rationale comments now describe the code beneath them (#297)

### LINKED ISSUE #297: Two comments in the registry stack describe behavior the code no longer has
Two comments in the registry stack describe behavior that is not the behavior
of the code beneath them. Both are the rationale comment for the thing they sit
on, which is the worst place for this: they are the first thing a reader trusts
and the last thing they re-check.

## 1. `turn-derivation.js` documents the branch the hardening deleted

`clients/web/turn-derivation.js:17-25`, the header on the `CLOCK_CODES` table:

> Stamps arrive in two dialects: wire-decoded stamps already carry the numeric
> u8 clock code and pass through untouched; script-built stamps name their clock
> and map through this table. **An unknown name maps to 0, which the core treats
> as its own stream key.**

`clockCode` immediately below it (`:30-33`) returns `null` for an unknown name,
and `update` (`:85`) turns `null` into a reset of the derivation. The
"maps to 0" path was removed precisely because collapsing distinct unknown
clocks onto key 0 bridges two streams as one — that is the defect luofang34/Pilotage#290 fixed,
and it is pinned by `two different unknown clocks never collapse into one
stream` in `turn-derivation.test.mjs`.

The replacement comment at `:27-29` is correct. The stale sentence above it
should go, and the surrounding block reworded so it does not read as if the
table has a fallback.

## 2. `descriptors.rs` module doc is future-tense about work that landed

`crates/pilotage-instrument-panels/src/descriptors.rs:4-9`:

> The masks and identities that **used to live** as shell constants are owned
> here, beside the panels they describe. `group_regions` and `extreme_states`
> **are populated when their consumers land** (the admission harness and the
> canonical-state corpus); `raster_baseline` **is populated when** the pinned
> frame hashes travel into the descriptors.

The consumers landed. `group_regions` is populated for all three panels
(`:77`, `:176`, `:264`) and `raster_baseline` for all three (`:149`, `:230`,
`:274`). The doc still promises them as future work. (`extreme_states` really
is still empty everywhere — that is its own gap, tracked separately.)

This trips two of the repo's comment rules at once: "used to live" is temporal
phrasing whose home is the commit message, and "populated when their consumers
land" is an annotation whose truth depended on a future event that has since
happened. The module doc should say what the descriptors are, not how they were
assembled.

## Ask

Doc-only fix, both files in one pass. No behavior change, nothing to pin — the
tests that make the `turn-derivation.js` sentence false already exist and pass.

Refs luofang34/Indicate#36 luofang34/Pilotage#285 luofang34/Pilotage#286 luofang34/Pilotage#287 luofang34/Pilotage#290.

luofang34:
The CLOCK_CODES header no longer claims an unknown clock maps to key 0 — clockCode refuses unknowns precisely because collapsing distinct unknown clocks onto one key would bridge two streams, and the tests pin that. The descriptors module doc says what the descriptors are instead of how they were assembled; the populated fields speak for themselves.

Closes #297.

---
- <kbd>&nbsp;2&nbsp;</kbd> #301 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #300


luofang34:
Independent review disposition: the CLOCK_CODES header promised refusal the lookup did not enforce — `CLOCK_CODES[clock] ?? null` resolves prototype keys (constructor, __proto__, toString) to non-null values that then marshal into the wasm call. clockCode now uses Object.hasOwn, and prototype-key refusal is pinned by tests. The turn-derivation size cap (92) rides this PR per the ratchet discipline, and the descriptors module doc now includes extreme states in the contract enumeration.

PR #302, opened 2026-08-06
Quat::from_euler: the convention owner carries both directions (#294)

### LINKED ISSUE #294: pilotage-frames has to_euler but no inverse, so every attitude producer writes its own
`pilotage-frames` publishes `Quat::to_euler` but no inverse, so anything that
needs to *build* an attitude from roll/pitch/yaw writes the conversion itself.

Two copies now exist downstream in the avionics-link prototype: one drives the
display self-test with a known attitude, and one feeds live aircraft attitude
from a panel's AHRS stream. They are identical today and there is nothing to
keep them that way. The crates they live in must not depend on each other — one
is a link bridge, the other a panel host, and they meet only at the packed state
block — so neither can borrow the other's copy.

Hand-written rotation maths is exactly the kind of thing that is wrong in one
axis and looks entirely plausible, which is why both copies are round-tripped
against `to_euler` rather than trusted. That is a test for a problem that should
not exist: the crate that owns the convention should own both directions of it.

## Ask

`Quat::from_euler(roll, pitch, yaw)` — the aerospace ZYX inverse of the existing
`to_euler` — with the round-trip pinned where the convention lives. Every
downstream copy then deletes.

Worth doing before the panel work in luofang34/Indicate#36 goes further: a reusable panel that
needs to construct an attitude would otherwise make a third copy, and a
third-party panel would make a fourth.

luofang34:
The aerospace ZYX inverse of to_euler — yaw about Z, then pitch about Y, then roll about X, same signs — so no attitude producer hand-writes the conversion again. The round trip is pinned where the convention lives, alongside unit-norm and single-axis equivalence against the axis-angle constructor the crate composes from. Downstream copies can now delete.

Closes #294.

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #304
- <kbd>&nbsp;2&nbsp;</kbd> #303
- <kbd>&nbsp;1&nbsp;</kbd> #302 👈


luofang34:
Independent review disposition: the doc claim that no producer hand-writes the conversion was premature — three test-local copies survive (two of them evidence-bound locators). Comments reworded to present-tense intent; retiring the copies is #307 (needs an instruments-evidence re-record, so it gets its own PR).

PR #303, opened 2026-08-06
Admission verifies BackgroundCapability in both directions (#295)

### LINKED ISSUE #295: A panel's BackgroundCapability is declared but never verified — the #289 defect passes admission today
A panel's `BackgroundCapability` is a declaration nothing checks. The admission
harness validates that every band in `required_layers` is present
(`admission.rs:300`, `required_layers & !layers.present`), and never reads
`panel.background` at all — the only occurrences of the field anywhere in
`pilotage-instrument-conformance` are descriptor initializers in
`admission/tests.rs`.

So the capability is load-bearing for a compositor and unverified by the thing
whose job is admitting panels.

## This is not hypothetical — it already shipped

The monitor panel landed in luofang34/Pilotage#289 declaring `NotUsed` while painting an opaque
black rectangle. CI was green, the panel passed all 76 admission cases, and the
scene digest was pinned over the wrong declaration. It was corrected in luofang34/Pilotage#291 by
human review, not by a gate.

Reintroducing exactly that defect still passes today. Flipping
`MONITOR_DESCRIPTOR.background` back to `NotUsed` while leaving the opaque
`rect` in the `Background` band:

```
test admission::tests::a_panel_missing_its_required_band_is_refused ... ok
test admission::tests::an_always_fabricating_panel_is_refused_at_the_furniture_stage ... ok
test admission::tests::a_panel_leaking_numbers_into_a_withheld_region_is_refused ... ok
test admission::tests::a_transform_cannot_hide_a_dishonest_numeral ... ok
test admission::tests::builtin_panels_pass_admission ... ok

test result: ok. 5 passed; 0 failed
```

The only failure is `the_builtin_scene_digest_is_pinned`, and it is the wrong
instrument for this: it fails as an opaque hash mismatch with no diagnostic, and
it proves the declaration *did not change* — not that it is *true*. Had the
monitor's original `NotUsed` been correct-looking enough to survive review, the
pin would simply have been set over the false value and stayed green forever.
That is what happened between luofang34/Pilotage#289 and luofang34/Pilotage#291.

## Both directions are unguarded

- **Over-declares (`NotUsed`, paints anyway).** A compositor told the band is
  not painted allocates or composites something else there, and gets a black
  rectangle over it. This is the luofang34/Pilotage#289 case.
- **Under-declares (`Opaque`, paints nothing).** A compositor told the panel
  owns an opaque ground skips its own fill and gets transparent pixels where it
  was promised coverage. Nothing catches this either.

The second direction matters more for a third-party panel, which is the whole
premise of luofang34/Pilotage#267 — foreign panels admitted by conformance, not trust. A panel
author who declares `Opaque` out of optimism is admitted.

## It now propagates into the cross-shell contract

`digest.rs:84-88` binds `background` into `digest_panel_contract`, so the
declared value is part of the digest every shell must reproduce. An unverified
declaration is now baked into the contract that is supposed to prove two shells
show the same instruments.

## Ask

A fifth admission family, beside the layer contract, budgets, glyph coverage,
and honest status: the background capability must match what the panel emits.

- `Opaque` or `Cedeable` — the `Background` band is emitted under at least one
  canonical state.
- `NotUsed` — the `Background` band is never emitted, under any corpus or
  extreme state.

The inputs are already in hand: `draw_runs` decodes each scene, and
`validate_layers` already returns `present`. The check is a bitmask test per
case, and the `cfg(test)` fixture pattern from luofang34/Pilotage#267 gives it a panel that must
fail — a `NotUsed` descriptor over a ground-painting draw function.

Refs luofang34/Pilotage#267 luofang34/Indicate#36 luofang34/Pilotage#289 luofang34/Pilotage#291.

luofang34:
The declared capability was load-bearing for a compositor and checked by nothing — the shipped NotUsed-while-painting defect passed every admission case and only moved an opaque digest pin, which proves a declaration did not change, never that it is true. The harness gains a fifth family: NotUsed may never open the Background band, and an owned band (Opaque or Cedeable) must carry a full-frame opaque paint, tracked through the transform state like the text walker. Cedeable's ceding under configuration stays pinned by the panel's own byte-equivalence tests; the harness draws the empty config, so it verifies the band-owning default.

Both defect directions are pinned: the shipped NotUsed-painter class and the optimist that declares Opaque over an empty band are refused with the declaration and the defect named. The shipped panels pass.

Closes #295.

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #304
- <kbd>&nbsp;2&nbsp;</kbd> #303 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #302


luofang34:
Independent review disposition: the coverage scanner was rewritten as a real graphics-state machine — Save/Restore push/pop transform+clip+alpha together, ClipRect is mapped and intersected and must contain the frame, coverage requires an axis-aligned CTM (bbox containment is exact, not conservative), and NotUsed now fails only on actual paint commands (opening the band empty is tolerated). The Rect-only coverage rule is stated in the module doc and in the refusal text. Three verified evasions (clip, rotate, alpha-restore) are pinned as refusal fixtures; the family count wording now says five everywhere.

PR #304, opened 2026-08-06
Every shipped panel declares its own hardest cases (#298)

### LINKED ISSUE #298: extreme_states is plumbed through admission and the digest, and empty on every shipped panel
`extreme_states` is wired end to end and used by nobody.

The mechanism is complete:

- `descriptor.rs:110` declares the field.
- `registry.rs:194` validates each entry's id and rejects duplicates.
- `admission.rs:199` chains them onto `CANONICAL_STATES`, so each contributes a
  full row of the withholding matrix.
- `digest.rs:67` folds them into the cross-shell scene digest.
- `registry/tests.rs:177` exercises the plumbing with a synthetic descriptor.

And every shipped panel declares `extreme_states: &[]` —
`descriptors.rs:145` (PFD), `:229` (HSI), `:273` (monitor).

## Why this is worth an issue rather than a shrug

The capability is stated as delivered. luofang34/Pilotage#286: "drives every registered panel
through the shared corpus **plus its own extreme states**", and "Panels
contribute stress fixtures through their descriptors." luofang34/Pilotage#266 makes the parallel
argument for `raster_baseline` and follows through — a panel travels with its
own regression baseline, and `all_builtin_panels_carry_a_baseline` keeps a
shipped panel from ever losing one. `extreme_states` got the same argument and
no follow-through, and has no equivalent floor test.

The gap matters most for the case the harness exists to serve. The shared
corpus is the *common* situations — nothing-fed, typical, fully-fed,
source-unusable. A panel's extreme states are where its own arithmetic breaks:
the PFD past +/-90 degrees pitch and across the 360/0 heading wrap, the HSI with
a course exactly reciprocal to track or a zero-distance waypoint, the monitor
with eight maximum-length lines against the frame budget and the glyph
vocabulary. None of those are reachable from the corpus, so no admission case
covers them, and the frame-contract debt counted as warnings (the pitch ladder's
144) is measured only over the gentle states.

## Ask

Either:

1. **Populate them.** Each shipped panel declares the states where its own
   drawing is hardest, and a floor test in the shape of
   `all_builtin_panels_carry_a_baseline` keeps a shipped panel from declaring
   none. This moves the scene digest once, for a stated reason, as the contract
   requires.

2. **Or scope the claim.** If per-panel stress fixtures are deliberately later
   work, say so where the capability is described (the `#[doc]` on the field
   and the conformance crate's module doc) so the harness is not read as
   covering ground it does not.

(1) is the one that buys something. Filing this rather than fixing inline
because choosing each panel's extreme states is a display-engineering judgement,
not a mechanical fill-in.

Refs luofang34/Pilotage#266 luofang34/Pilotage#267 luofang34/Indicate#36 luofang34/Pilotage#286.

luofang34:
The extreme-state mechanism was wired end to end and used by nobody. The PFD gains an inverted, hard-rolling unusual-attitude state and a readout-extremes state (wide and negative values in every box, heading on the 360/0 wrap); the HSI a reciprocal-course state (full-scale deviation, zero-distance waypoint, wrapped heading); the monitor a full channel of eight maximum-length lines. A floor test in the shape of all_builtin_panels_carry_a_baseline keeps a shipped panel from declaring none. The admission matrix grows to 98 cases and the digest moves once, because the corpus each panel must reproduce grew.

The new states immediately earned their keep: at full-scale vertical speed the VSI label slid its ink under the baro and selected-altitude boxes (its clamp now parks it clear of both rows — typical values are untouched, so the pinned frame hashes hold), and a maximum-length monitor line at the old text size overran the frame edge (rows now size to fit the frame from the margin; the monitor baseline re-pins with the digest). The PFD selected-altitude honest-status region is deliberately withdrawn with its reason recorded: the altitude tape's scale ladder legitimately runs ink through that area at extremes, and the HSI's heading-select region carries the selections coverage.

Closes #298.

---
This is **part 3 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #304 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #303
- <kbd>&nbsp;1&nbsp;</kbd> #302


luofang34:
Independent review disposition: (1) the selected-altitude withdrawal rationale claimed occlusion, but the honest-status check runs with Selections withheld — the box is absent in the failing case; reworded to the real reason (ladder-label y moves with altitude, no region geometry separates the two). (2) A second independent pass found the reciprocal-course fixture's course was reciprocal to the inherited typical heading, not the wrapped heading the fixture itself sets — course_rad is now exactly pi from the in-state heading and the scene digest re-pins once (850f9298…); 98 cases and the 59-warning ratchet are unchanged. (3) The monitor stress fixture's rows are now pinned at full capacity so the TextLine fallback can never silently hollow the fixture.

PR #305, opened 2026-08-06
The ABI golden fixtures live inside the crate that owns the codec (#262)

luofang34:
Three of the five out-of-closure references blocking extraction were the v6 golden hex frames: the state crate reached ../../clients/web/fixtures for its own codec's fixtures. They now live in crates/pilotage-instrument-state/fixtures/, so they travel with the crate at the split; the JS golden test and the xtask regenerator point at the codec owner's tree, and the regenerated bytes are identical. The remaining two out-of-closure references — the scene-conformance corpus and the REN-04 timing artifact — are the corpus-ownership decision luofang34/Indicate#35 tracks separately.

Refs luofang34/Indicate#35 luofang34/Indicate#36.


luofang34:
Independent review disposition: six in-tree pointers still named clients/web/fixtures (xtask help text, the create_dir error context, three module docs, the JS writer header) — all now name the codec-owned location. `grep -rn 'clients/web/fixtures'` over the tree returns nothing.

PR #306, opened 2026-08-06
Validate the shell composition once, not twice per panel per frame (#296)

### LINKED ISSUE #296: The shell re-validates the whole registry composition twice per panel per frame
`panel_registry.rs:17` builds a fresh `Registry` on every call:

```rust
pub(crate) fn registry() -> Option<Registry> {
    Registry::new(BUILTIN_PANELS).ok()
}

pub(crate) fn descriptor(panel: u32) -> Option<&'static PanelDescriptor> {
    registry()?.panels().get(panel as usize)
}
```

`Registry::new` is not a wrapper — it runs the full composition validation
ladder (`registry.rs:150-205`) across every panel: id well-formedness, the
duplicate-id scan, layer bits against `DEFINED_LAYER_BITS`, design-frame
finiteness, strictly-ascending schema keys, frame containment for every
`group_regions` entry, and the duplicate-extreme-id scan.

The render path calls `descriptor()` twice per panel per frame:

- `exports.rs:160` — `render_into` resolves the descriptor to get `draw`.
- `exports.rs:120` — `validate_panel_scene`, reached from
  `validate_and_commit_scene` at the end of the same `render_into` call,
  resolves it again to get `required_layers`.

At three panels and a ~50 Hz presentation that is on the order of 300 full
composition validations per second, in wasm, over a `&'static` input whose
answer cannot change between calls.

## Why it is worth fixing now rather than when it hurts

It is cheap today and the profile will not show it. The problem is the shape:
the cost is O(panels x regions per panel), and the entire premise of luofang34/Indicate#36 is
that panels get added — "adding a panel is a registry entry, nothing else"
(#289). Every panel added multiplies work already being done redundantly, in
the one path that runs every frame. The bench shell (#288) and a future iPad
shell inherit the same accessor.

The enumeration exports (`panel_count`, `panel_id`, `panel_title`, ...) each
re-validate too, but those are load-time and do not matter.

## Ask

Validate the composition once and hand out the result. `BUILTIN_PANELS` is
`&'static` and `Registry::new` is pure — no I/O, no environment — so a
`OnceCell`/`OnceLock` here does not run into the module-level-state rule. wasm
is single-threaded, so a `thread_local!` is also available if that reads
better.

The failure semantics should not change: a composition that does not validate
must still surface as `RenderStatus::InvalidPanel`, fail-visible, exactly as
`registry()` returning `None` does today.

Refs luofang34/Indicate#36 luofang34/Pilotage#287 luofang34/Pilotage#288.

luofang34:
registry() memoizes the validated composition in a OnceLock: the input is &'static, Registry::new is pure, and the render path resolves a descriptor twice per panel per frame — O(panels × regions) work whose answer cannot change, in the one path every added panel multiplies. Failure semantics unchanged: a composition that does not validate still memoizes as None and surfaces as InvalidPanel, fail-visible.

Closes #296.

### DIFF
--- a/clients/web-instruments/src/panel_registry.rs
+++ b/clients/web-instruments/src/panel_registry.rs
@@ -13,9 +13,13 @@ use pilotage_instrument_registry::{
 use wasm_bindgen::prelude::wasm_bindgen;
 
 /// The validated shell composition, or `None` if the shipped panels no
-/// longer compose — which the panels crate's own tests make unreachable.
+/// longer compose — which the panels crate's own tests make
+/// unreachable. Validated once and memoized: the input is `&'static`
+/// and `Registry::new` is pure, so the answer cannot change between
+/// calls, and the render path asks for it twice per panel per frame.
 pub(crate) fn registry() -> Option<Registry> {
-    Registry::new(BUILTIN_PANELS).ok()
+    static COMPOSED: std::sync::OnceLock<Option<Registry>> = std::sync::OnceLock::new();
+    *COMPOSED.get_or_init(|| Registry::new(BUILTIN_PANELS).ok())
 }
 
 pub(crate) fn descriptor(panel: u32) -> Option<&'static PanelDescriptor> {


luofang34:
Independent review disposition: no findings — memoization is pure over 'static input, failure semantics unchanged (a failed composition memoizes as None, fail-visible), and every hot-path consumer goes through registry().

PR #310, opened 2026-08-06
The instrument and geodesy suites delegate Euler conversion to the convention owner (#307)

### LINKED ISSUE #307: Retire the three test-local Euler→quaternion helpers in favor of Quat::from_euler
\`Quat::from_euler\` (#294 / PR #302) is the convention-owned inverse of \`to_euler\`, but three hand-written copies survive in test code:

- \`crates/pilotage-instrument-panels/src/pfd/attitude_tests.rs\` — \`quat_euler\`
- \`crates/pilotage-instrument-raster/tests/attitude_behavior.rs\` — \`euler_quat\`
- \`crates/pilotage-instrument-panels/src/hsi/tests.rs\` — \`quat_yaw\`

**Why they were not retired in #302:** the first two files are evidence-bound locators in \`docs/instruments/evidence-graph.evg\` (source-digests \`a836ec8a\` and \`7ab0d7f2\`). Editing them forces an instruments-evidence re-record (re-run suites, re-digest, re-anchor), which is out of scope for a comment-accuracy fix and deserves its own reviewed PR.

**Acceptance:**
- All three helpers deleted; call sites use \`pilotage_frames::Quat::from_euler\`.
- Frame hashes / scene digest unchanged (the helpers are numerically identical — the round-trip test in \`rotation/tests.rs\` discriminates axis order with all-three-angles-nonzero triples).
- Instruments evidence graph re-recorded for the two bound files; evidence gate green with \`--require-resolvable\`.

luofang34:
Retires the hand-rolled ZYX Euler-to-quaternion expansions within reach of `Quat::from_euler`: the three instrument-suite helpers (#307) and the geodesy conformal suite's copy — all four survive as one-line degree-taking adapters that delegate, so the call sites stay in degrees and no convention content remains in them. The remaining hand-written expansions (mavlink codec, session gate, aviate fixtures) build protocol or array types in crates with no frames dependency and stay.

Bit-identity holds because the adapters compute the same f32 operations: `to_radians` is a multiply by the same `PI/180` constant the deleted code used, halving by `/2.0` and `*0.5` are the same exact scaling, and `from_euler`'s four component expressions are term-for-term the deleted ones. The pinned conformal sim hash is an end-to-end witness on the geodesy side. (One caveat recorded: `from_euler(0,0,yaw)` produces `-0.0` in place of `0.0` in a zero component when `|yaw| > π` — value-identical, bit-different; no current fixture reaches it.)

The PFD and raster attitude suites are digest-bound evidence sources, so the ATT-01 results re-record over real runs in the companion evidence commit, anchored to the code commit; verified with the evidence gate: verdict VALID, `--require-resolvable` exit 0.

Closes #307.

https://claude.ai/code/session_013gPz9A5f9NbYFmop3UHMuw

### DIFF
--- a/crates/pilotage-instrument-panels/src/hsi/tests.rs
+++ b/crates/pilotage-instrument-panels/src/hsi/tests.rs
@@ -47,13 +47,7 @@ fn heading_only(heading_rad: f32) -> PanelData {
 }
 
 fn quat_yaw(yaw: f32) -> pilotage_instrument_state::Quat {
-    let h = yaw / 2.0;
-    pilotage_instrument_state::Quat {
-        w: libm::cosf(h),
-        x: 0.0,
-        y: 0.0,
-        z: libm::sinf(h),
-    }
+    pilotage_instrument_state::Quat::from_euler(0.0, 0.0, yaw)
 }
 
 fn render(data: &PanelData) -> Vec<u8> {
--- a/crates/pilotage-instrument-panels/src/pfd/attitude_tests.rs
+++ b/crates/pilotage-instrument-panels/src/pfd/attitude_tests.rs
@@ -13,17 +13,11 @@ use super::{PfdConfig, VSpeeds};
 
 /// f32 ZYX euler → quaternion for orientation fixtures.
 fn quat_euler(roll_deg: f32, pitch_deg: f32, yaw_deg: f32) -> Quat {
-    let d = core::f32::consts::PI / 180.0;
-    let (r, p, y) = (roll_deg * d / 2.0, pitch_deg * d / 2.0, yaw_deg * d / 2.0);
-    let (cr, sr) = (libm::cosf(r), libm::sinf(r));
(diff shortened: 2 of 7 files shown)


luofang34:
Independent review disposition: MERGEABLE with wording corrections. The overclaimed subject ('the only Euler conversion left standing') is resolved by actually retiring the geodesy suite's copy in a follow-up commit — the surviving expansions (mavlink, session, aviate fixtures) have no frames dependency and legitimately stay. The 'proven by frame hashes' citation was unfounded (those paths never call the helpers); the PR body now states the real argument (identical f32 operations term-for-term) plus the reviewer's empirical witness (914M-triple bit comparison, zero mismatches) and records the -0.0 caveat past |yaw|>π. Per the reviewer's constraint, commit 0fd556b was never amended — it is the evidence anchor; the corrections landed as an appended commit and a PR-body rewrite.

PR #311, opened 2026-08-06
Honest status becomes a provenance rule: numerals claim the group that fed them (#309)

### LINKED ISSUE #309: Honest-status is positional, so shared screen space forces a coverage loss — attribute numerals to the group that fed them
## The immediate regression

luofang34/Pilotage#304 removed the PFD's `Selections` honest-status region. The PFD's
selected-altitude readout — the only thing on that panel fed by the `Selections`
group — is now unguarded.

Verified, not inferred. Patching `baro_and_sel_boxes` so the `(_, None)` arm
paints a plausible number in the readout slot — i.e. a panel that fabricates a
selected altitude while `Selections` is withheld:

```rust
(_, None) => {
    scene.fill_color(palette::CYAN)?;
    scene.text(435.0, 12.0, 18.0, Anchor::CENTER, "10300")?;
}
```

```
test admission::tests::builtin_panels_pass_admission ... ok
test result: ok. 1 passed; 0 failed
```

That case was refused before luofang34/Pilotage#304. The panel's behaviour is still honest today —
`baro_and_sel_boxes` paints the box only when `altitude_sel_m` is `Some` — so
this is a lost guard, not a live defect.

**The removal was justified and should not simply be reverted.** The altitude
tape's scale ladder is `Kinematics` ink: labels anchor at `x=408`, size 18, at
`y = CENTER_Y - (ft - alt) * PX_PER_FT`, clipped to `(390, 0, 90, 335)`. Their y
moves with altitude, so at some altitudes a ladder label sits inside the old
`(390, 0, 90, 24)` region and `check_honest_status` refuses honest scale ink.
There is no x separation either — the readout anchors CENTER at `x=435`, inside
the ladder's band. The new `pfd_readout_extremes` state reaches those altitudes,
which is how the conflict surfaced. Restoring the region under the current model
reintroduces a false positive.

## The class problem

The harness's honest-status model is *"this screen box belongs to one group."*
Instruments do not work that way: two sources legitimately share screen space,
and the altitude tape is simply the first place it bit. Any future panel that
overlays a readout on a moving scale hits the same wall, and the only moves
available under the current model are to delete the region (lose coverage, as
here) or to shrink it until it no longer covers the readout (same thing, slower).

There is a second, quieter weakness in the same model: **only declared regions
are policed at all.** A fabricated numeral drawn anywhere outside a declared
region is invisible to the harness today, on every panel.

## The durable fix: attribute a numeral to the group that fed it

Tag each numeric text run with the state group its value derives from, and
replace the positional rule with a provenance rule:

> A text run tagged `G` must not be visible when `G` is withheld.

This fixes the class rather than the instance:

- Honest `Kinematics` ladder labels pass through the selected-altitude box
  freely — position stops mattering.
- A fabricated `Selections` number is refused wherever it is drawn.
- Coverage extends to every numeral on the panel, not only those inside a
  declared region.

**It stays adversarial — the tag cannot be gamed by simply lying.** A fabricator
that mislabels its fake selection number as `Kinematics` is then refused when
*Kinematics* is withheld, because the fabricated number is still drawn. The tag
is a claim the withholding matrix immediately tests. The one residual escape is a
fabricator whose fake value is gated on the group it falsely claims — at which
point it is, behaviourally, deriving from that group. Worth stating in the design
rather than overclaiming soundness.

Two properties this needs to be sound:

1. **Tags must be total.** If a numeric run may be untagged, a fabricator omits
   the tag and escapes everything. Every numeric text run carries a group, and an
   untagged numeral is itself an admission failure.
2. **Tags supersede regions for this property.** `group_regions` would no longer
   drive `check_honest_status`. It stays useful for the separate
   "the readout slot must visibly dash out" property, which is a different check
   and worth keeping distinct.

## Cost — lower than it looks, because the format was built for this

The scene IR encodes every command as `[opcode u8][payload_len u16 LE][payload]`,
with the explicit length present precisely so decoders can skip opcodes they do
not know; `Cmd::Unknown { opcode }` exists and `layer.rs` counts unknown opcodes
as skips. So an appended attribution opcode is **forward-compatible by
construction**: the browser interpreter, the reference rasterizer, and the
CoreGraphics backend all skip and count it with no change.

- **No corpus bump is forced.** `scene-conformance-corpus.json` is a
  generator/hand-authored set of byte traces (`bytesHex`), not panel output — what
  panels emit does not touch it. A bump is only wanted later, deliberately, to
  give the new opcode its own coverage entries.
- **Budget headroom is ample.** `MAX_SCENE_BYTES` is 64 KiB and
  `MAX_LAYER_COMMANDS` is 4096 per layer; attribution costs ~4 bytes per
  numeric run. The constraint worth watching is the *command* budget in `Tapes`,
  where the ladder labels concentrate — which argues for a prefix form
  ("the next text run is group G", 1 command per run) over a bracket form
  (Begin/End, 2 commands per run).
- **The scene digest and the raster baselines move once**, for a stated format
  reason, as the contract requires.
- `SCENE_FORMAT_VERSION` does not need to change — appending an opcode is the
  documented growth path; only layer-id additions require a version bump.

## The cheaper alternative, and why it is weaker

Without any format change: keep the region, and refuse a numeral in `region(G)`
under withholding-of-`G` **only if it survives withholding of every other
required group**. The ladder label vanishes when `Kinematics` is withheld and so
passes; a fabricated constant survives everything and is refused. This needs no
tags, no opcode, no digest move — it is pure harness-side computation over draws
the matrix already performs.

It is weaker in two ways: a fabricator whose fake number is gated on any
unrelated fed group escapes it, and it leaves the "only declared regions are
policed" hole untouched. Reasonable as a stopgap that restores the lost coverage
this week; not the thing to build on.

## Ask

Decide between the two. If attribution is the direction, the shape is: append the
opcode in `pilotage-instrument-scene`, thread a tag through `SceneWriter::text`
for numeric runs, make totality an admission family, and retire
`check_honest_status`'s positional rule — at which point the PFD's
selected-altitude coverage comes back for free and the ladder conflict cannot
recur.

Until either lands, the `Selections` gap on the PFD is real and the descriptor
comment should say *that*, rather than pointing at the HSI's heading-select
region — which is a different panel and a different readout, and does not cover
this one.

Refs luofang34/Pilotage#267 luofang34/Indicate#36 luofang34/Pilotage#295 luofang34/Pilotage#298 luofang34/Pilotage#304. Blocks nothing, but luofang34/Indicate#35/#268 both treat the
admission harness as the gate that makes an out-of-repo panel safe, so the model
is better settled before third-party panels depend on it.

luofang34:
The positional model — one screen box, one group — lost the PFD selected-altitude coverage the moment two sources legitimately shared the altitude strip, and it never policed a numeral outside a declared region. The scene IR gains an appended Attribute opcode (0x31, prefix form, atomic with its run via SceneWriter::text_attributed; no format-version bump — appended opcodes are the documented growth path); every backend knows it as a paints-nothing command so the version-skew skip counters stay quiet.

The admission harness replaces the region rule with three provenance rules, adversarially closed: totality (a numeric run with no claim is refused outright — omission is not an escape), bounding (a claim outside the panel's required groups is refused — the matrix could never test it, and the empty-config harness refuses configuration claims the same way), and the claim test itself (a run claiming the withheld group may not be visible, wherever it is drawn). The test is against withholding, not the resolved group summary — a per-field signal may legitimately outlive its group status (the baro setting is not an estimate and survives absent trust), and the resolve layer's per-field gates remain the authority on what may be shown. The scanner pairs claims to runs and refuses dangling or shape-interposed claims at the byte level.

Every builtin numeral now carries its claim (ladders, readouts, boxes, monitor lines, source labels through the shared symbology helpers), the PFD selected-altitude coverage returns as the claim on the run itself, and group_regions remain the descriptor's readout-ownership statement rather than the numeral police. The matrix stays at 98 cases and the 59-warning ratchet holds; the scene digest re-pins once for the stated format reason; raster pixel baselines are untouched because attribution draws nothing.

Closes #309.

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #313
- <kbd>&nbsp;2&nbsp;</kbd> #312
- <kbd>&nbsp;1&nbsp;</kbd> #311 👈


luofang34:
Independent review disposition: MERGEABLE; all five minors and three nits addressed. The scope block now states the withholding-not-resolved-status limitation, the unexercised source-label claims (stateful source-monitor gate), and the empty-config consequence for config-driven foreign numerals (also stated at the refusal site). The altitude ladder and readout claims now follow the declared altitude class — kinematic for local-relative, Air otherwise — so a barometric altitude cannot be falsely refused under kinematics withholding when a non-default class becomes reachable (byte-neutral for the shipped corpus, digest unchanged). The orphaned scanner paragraph moved to the background-checks module it describes; the atomic-pair wording now says what the writer guarantees; 0x31 joins the JS known-no-op test so a dropped case cannot silently inflate the version-skew counter; group_regions' field doc says plainly that no shipped consumer reads it; and both dash paths carry the WHY for staying unclaimed.

PR #312, opened 2026-08-06
The HSI presents track-up for heading-less sources, annunciated TRK (#260)

### LINKED ISSUE #260: The HSI is inert for any source without magnetic heading (ADR-0026 data-gateway profile)
The HSI gates its entire rose on a valid heading:

```rust
// crates/pilotage-instrument-panels/src/hsi.rs
let hdg = data.heading.value_rad;
if hdg.status.shows_value() {
    rose::draw_rose(scene, hdg.value)?;
    if data.heading_bug_rose_rad.status.shows_value() { … }
    if data.track_rad.status.shows_value() {
        rose::draw_track_diamond(scene, hdg.value, data.track_rad.value)?;
    }
}
```

The track diamond is positioned *relative to* heading, so it is correctly
unavailable without one. Guidance is gated the same way.

## Why this matters now

ADR-0026's `data-gateway` profile is "a bridge to an AHRS/GNSS/ADS-B receiver,
including flight-plan send/receive where the device supports it". A real
instance of that profile — a certified GPS navigator bridged over its serial
protocol — publishes GNSS position, pressure altitude, traffic, and a flight
plan, and **no magnetic heading at all**. It does publish ground track.

Against such a source the HSI can therefore only ever paint its background, the
flagged heading box, four dashed data boxes, and the failure X. Measured
concretely: the encoded scene is byte-identical whether the source is fed or
completely unfed (561 bytes either way), while the PFD changes with its input
(884 unfed, 1298 fed). The HSI is structurally inert for this entire profile.

## What this issue is not asking for

Not deriving heading from track. They differ by drift angle in any wind, and
substituting one for the other is exactly the "plausible number" the validity
model exists to prevent. A consumer doing that silently would be worse than a
dark rose.

## The question

Whether the HSI should support a track-referenced presentation for sources that
have no heading — the way a GPS-only navigator annunciates a track-up rose —
and if so how the reference is annunciated so it can never be mistaken for
heading. `HeadingReference` currently offers `Magnetic`, `True`, `SimLocalTrue`,
and `Unknown`; there is no track option, and adding one touches the state
vocabulary, the panel, and the `HDG REF` annunciator together.

This is the display half of the same gap as #256: a source whose group set
differs from a flight controller's is representable in principle but has no
panel that can usefully render it.

## Evidence

Found while bringing up a Core Graphics backend for the scene IR (#255) against
a real panel link. The backend agrees with the reference rasterizer on 99.8% of
lit pixels for both panels, so this is a data-availability limit rather than a
rendering fault.

luofang34:
The data-gateway profile — a certified GPS navigator publishing position, track, and guidance with no magnetic heading — left the HSI structurally inert: byte-identical scenes fed or unfed. The resolve layer now selects a RoseBasis (heading sample when it shows, else the NED ground track, else unavailable) and derives the display reference from it, so every converted angular quantity — wind, bug, course — presents against the same reference the rose is oriented by and the panel can never disagree with its own conversions. Under the track basis the display reference is True, the frame a NED track natively lives in; a magnetic course without usable variation fails conversion and draws no CDI, so references never mix on one rose.

This is a presentation, not a substitution: track is never offered as heading. The value in the digital box wears the selection color under a track basis and the reference slot annunciates TRK — no heading reference label can appear on a track rose — the rotating symbology and the value box claim Kinematics for the provenance family, and the panel gates every paint on the signal's own status as well as the basis. A stationary source (below the track groundspeed floor) fails visibly exactly as before.

One shipped corpus state changes behavior, disclosed and pinned: with a present-but-untrusted heading (`source-unusable`) there is now NO display reference, so the heading-select bug no longer presents — the select box fails closed to its label instead of showing a selection against a dead rose. A regression test pins the closed behavior; the digest re-pin carries the byte change.

The HSI gains a track-up extreme state, so the mode is admitted and withheld like every other corpus case: 106 cases, ratchet unchanged.

Closes #260.

https://claude.ai/code/session_013gPz9A5f9NbYFmop3UHMuw


luofang34:
Independent review disposition: the MAJOR is fixed — the commit's 'every previously pinned scene is byte-identical' claim was false (hsi × source-unusable shrinks: with a present-but-untrusted heading there is no display reference, so the heading-select bug fails closed to its label). The PR body now discloses the change and why it is an improvement, and a regression test pins the closed behavior. The TRK/TRU distinctness minor: the track value itself now wears the selection color in the box, so the readout cannot read as a heading before the eye reaches the annunciation. REF joins the reference-label exclusion so the test covers the full vocabulary.

PR #313, opened 2026-08-06
Flight director: the group, the bars, and the three recorded decisions (#261)

### LINKED ISSUE #261: No flight director: neither command bars on the PFD nor a guidance-command group in state
The PFD draws no flight-director command bars, and no state carries them.

Searching the panels crate for a director turns up exactly one hit, and it is a
disclaimer rather than an implementation
(`crates/pilotage-instrument-panels/src/pfd.rs:148`): the element it describes is
"an orientation cue, never a flight-director command".

`AircraftState` has no guidance-command group either. `NavData` carries course,
lateral deviation in dots, optional vertical deviation, TO/FROM, and distance —
deviation from a path, not a commanded attitude to fly. A director needs
commanded pitch and roll, plus the mode and engagement state that say what is
commanding them.

## Why it is worth recording now rather than discovering later

Three properties of the existing design make a director more than a pair of
extra bars, and all three want deciding before something is drawn:

- **Criticality placement.** Director bars are guidance, so they belong in
  `LayerId::Guidance` — but they overlay the attitude field, which is
  `LayerId::Attitude`. Encoding order is z-order and layers appear once each in
  ascending order, so bars drawn in Guidance paint over the horizon
  correctly. That works, but it should be an explicit decision recorded with
  the layer contract rather than an accident of which layer someone picked.

- **Validity.** A director must never show a command derived from a degraded or
  stale source, and the existing `Trust::fold` discipline gives the mechanism.
  Directors also have to disappear cleanly rather than freeze, which is a
  stronger requirement than the dashes used for a readout.

- **Mode annunciation.** A command bar without the mode that produced it is not
  interpretable, so the annunciation is part of the feature, not a follow-up.

## Not asking for

A director synthesised on the display side from deviation. Turning cross-track
error into a command bar is an autopilot function, and inventing one in the
presentation layer is the same class of mistake as deriving heading from ground
track (#260).

## Context

Raised while driving the panels from a real avionics link and building a
preflight display check for them, where the operator confirms each element
against a stated value. The director is the one element on that card that could
not be written, because nothing upstream can produce it.

luofang34:
The state vocabulary gains the FlightDirector group (wire tag 0x0D; the never-assigned planned reservations shift down one — an allocation is a doc edit before it is a variant): commanded pitch and roll with the mode and engagement that say what is commanding them, because a command bar without its mode is not interpretable. The add-a-group recipe carries it end to end — type file, GroupId variant, the four exhaustive matches, validate envelope (unknown mode or engagement is a fail-closed sentinel; commands beyond the attitude envelope fault the group), resolve fold, and the JS emitter forced by the regenerated golden frames. Nothing synthesises a director from deviation: the display draws only what an upstream autopilot commands.

The three decisions the issue asked to record, recorded where they bind: (1) criticality placement — bars draw in the Guidance band, which overlays the Attitude field because layers encode once each in ascending z-order; the band is always emitted so the layer contract holds under every degradation. (2) validity — bars draw only from a fully valid, ENGAGED director over a shown attitude, and under any degradation they disappear entirely; a frozen or dashed command is still a command. (3) mode annunciation is part of the feature: FD ATT/NAV/APR annunciates when armed or engaged, claimed from the director group so the withholding matrix tests it.

The PFD requires the group and the Guidance band, and gains a director-engaged extreme state: 121 cases; the warning ratchet moves 59 to 83, all of it the same GS/SET readout-ink debt counted across the new state's nine cases. The digest re-pins once for the grown corpus; the PFD pixel baseline is untouched (typical feeds no director, and an empty band paints nothing).

Closes #261.

---
This is **part 3 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #313 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #312
- <kbd>&nbsp;1&nbsp;</kbd> #311


luofang34:
Independent review disposition: the MAJOR is fixed — the unusual-attitude declutter tier now strips the director bars like everything else that competes with the horizon (recovery is flown to the horizon, not to an autopilot command; the bars would overlay the chevron region from a higher band), pinned by a unit test pairing an engaged director with the unusual-inverted attitude. The pitch scale is now the horizon's own constant imported rather than a second literal. The fail-closed sentinels gained the per-branch tests the module doc advertises (unknown mode/engagement round-trip as 255; every director_fault branch has a validate_state case). ResolvedDirector's hand-written Default now agrees with what resolve emits for an absent group (Unknown). The type doc states plainly that no shipped posture publishes the group yet.

PR #314, opened 2026-08-06
The extraction boundary closes: owned artifacts, a standalone gate, a protocol-free shell (#262)

luofang34:
The last two reaches out of the instrument closure are gone: the scene-conformance corpus lives with the IR that defines its vocabulary (crates/pilotage-instrument-scene/corpus/ — the reference rasterizer reads a sibling crate, the browser suite reads into the crate tree and keeps its corpusSha256 pin as the cross-repo sync mechanism), and the REN-04 timing artifact lives with the raster crate whose tests consume it. Bytes unchanged in both moves.

The no_std CI gate splits: the closure compiles for thumbv7em in ONE standalone invocation containing only closure crates — the gate the extracted repository runs unchanged — with geo/ingress in their own step and the host-side conformance harness deliberately absent.

The browser wasm splits on the protocol dependency: pilotage-instruments-web-shell carries the instrument half (runtime, registry enumeration, feeder exports) with no protocol dependency by construction, pilotage-instruments-web keeps the wire/video decode half and bundles both into the unchanged cdylib — same artifact, same JS surface, and a CI step fails if the shell ever grows a protocol edge.

ADR-0034 (Proposed) records the boundary decisions extraction would otherwise freeze silently: the design-frame contract and its warning ratchet, artifact ownership (instruments evidence graph and AIR-* registry travel with the panels), gate placement, and rev-pin advance ownership keyed to the scene digest.

Refs luofang34/Indicate#35 luofang34/Indicate#36.


luofang34:
Independent review disposition: all three findings fixed. (1) detect-target.sh's dangling artifact path repointed. (2) The moved shell test is a byte-pure rename again; the Guidance-band fixture growth is its own commit stating the cross-change rationale (the fixture tracks the descriptor's required bands, which grow in the director change; an extra band is legal either way, so both merge orders pass). (3) The protocol-free CI guard is fail-closed: cargo tree runs as an assignment under set -e, so a failed tree aborts the step instead of reading as clean. Also applied: ADR-0034 breadcrumb anchors replace #-number markers in code/CI/crate metadata, the Decision-2 locator wording matches the tree, and the mechanically-substituted long lines are rewrapped.

PR #315, opened 2026-08-06
Evidence: re-anchor the ATT-01 baseline to the merged euler-retirement squash

luofang34:
The squash merge rewrote the lane commit CFG-WORKTREE declared, orphaning the baseline per the graph's merge/rebase rule. The config item re-anchors to the squash commit on main; the bound test sources are byte-identical, so the source digests are unchanged, and the artifacts re-stamp their config identity. Verified in a worktree over origin/main: verdict VALID, att01 suite green.

### DIFF
--- a/docs/instruments/evidence-artifacts/att01/panel.run.txt
+++ b/docs/instruments/evidence-artifacts/att01/panel.run.txt
@@ -2,9 +2,9 @@ Pilotage evidence — captured verification run (SIM / NOT FOR FLIGHT)
 scope: ATT-01
 suite: PFD attitude panel
 command: cargo test -p pilotage-instrument-panels
-config-digest: 0fd556b83e1304b3679c0f81e6ae61ba74311bb1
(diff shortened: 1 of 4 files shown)


luofang34:
Independent review disposition: the BLOCKER was real — the pushed commit was empty (the re-anchor hunks were dependency-locked to the still-applied euler-retirement lane, and the delivery tooling's 'could not be applied' note was filtered out of view). The branch now carries the full re-anchor exactly as the review specified: all seven lane-sha occurrences move to the squash c6a9e8aa, the CFG tree attr moves to the squash's tree, the three artifact headers re-stamp, and the three output digests match the reviewer's independently computed values. Rebased onto main so reachability is decidable; verified at the branch head: gate VALID with --require-resolvable, pilotage-evidence suite green. This PR must merge FIRST — main's hard att01 test is red until it lands, and every updated branch inherits that red.

Issue #321, opened 2026-08-08
[TRACKING] ARCH-01: composable aviation services across companion and iPad deployments


luofang34:
## Objective

Coordinate the source-neutral aviation services and operator-client modules
that Pilotage composes on a companion computer or iPadOS device.

This issue is a roadmap. Each implementation change uses a repository-local
issue and an independently reversible pull request.

**GitHub Project:** https://github.com/users/luofang34/projects/2

## Decision status

- [ADR-0036](https://github.com/luofang34/Pilotage/blob/main/docs/adr/0036-situational-domain-ownership.md)
  is `Accepted`. It supersedes ADR-0035.
- `docs/adr/0037-modular-operator-client-composition.md` defines
  operator-client composition from shared modules.
- The architecture records are the durable source of truth.
- This issue and the GitHub Project are execution and navigation surfaces.

ADR-0036 separates state by lifecycle. It does not select a repository or
process topology. ADR-0037 lets web and iPadOS clients use different user
interfaces over shared functional cores and data contracts.

## Proposed ownership

| Component | Owns | Does not own |
|---|---|---|
| `aero-link` | Radio access, framing, correction, protocol decode, bounded FIS-B assembly, and protocol classification | Weather lifecycle, notice lifecycle, tracks, or application consumer selection |
| `avionics-link` | Installed-avionics transport, protocol decode, and source adapters | A second domain state engine |
| Surveillance | Source-neutral traffic observations, fusion, freshness, tracks, deltas, and snapshots | Radio access, weather, aeronautical updates, or maps |
| Airmass | Weather observations, forecasts, hazards, revision, validity, expiry, and source comparison | Aeronautical updates, Navdata, or maps |
| `AeronauticalUpdates` | Perishable notices, cancellations, restrictions, operational status, validity, expiry, and source comparison | Weather, the Navdata baseline, link health, or maps |
| Navdata | The cycle-dated baseline for airports, runways, navaids, fixes, airspace volumes, airways, and procedures | Perishable weather or aeronautical updates |
| FlightPlanning | Drafts, Navdata resolution, static and regulatory checks, filing state, and immutable plan revisions | Activation checks, active-leg state, execution, or guidance |
| Briefing | One immutable point-in-time evidence result | Live domain state |
| Navigate | Ownship navigation, integrity, activation checks, active-plan state, execution, and guidance | Draft editing, filing state, actuation, or advisory state |
[...]

PR #330, opened 2026-08-09
Define lifecycle-owned situational domain boundaries

### LINKED ISSUE #348: Amend ADR-0036: prebaked baseline presentation, rendered cycle coherence, and the two time axes
Parent: luofang34/Pilotage#321

ADR-0036 is `Proposed` in #330. These three rules are cheap to add now and
awkward to add after acceptance. Each one elaborates a term that ADR-0036
already has.

## 1. The baseline is domain state. Only its presentation is prebaked.

The record must not let a reader think that the cycle-dated baseline is a map
asset. Navdata owns typed domain state, and FlightPlanning resolves against it.
The flight-plan handoff already carries the navigation-data snapshot ID, digest,
and cycle.

Add: a per-cycle offline build converts a Navdata snapshot into a versioned tile
bundle for a renderer. The bundle carries the cycle identity and the snapshot
digest. The bundle is an encoding at a renderer edge, in the same class as
GeoJSON. It is not domain state, and a domain must not produce it.

The reason for the rule: a full NASR extract must never enter the snapshot
path. A perishable update is a live feature. A baseline is a cached bundle.
Their cadences differ by six orders of magnitude.

## 2. A rendered composition states one cycle identity.

ADR-0036 defines a coherence requirement for a query. The same rule applies to a
rendered composition, and the record does not say so.

Add: a tile bundle and an overlay stream each carry a cycle identity. A
composition that mixes two cycle identities is invalid. A client rejects the mix
and reports it.

The reason for the rule: a runway designator changes when magnetic variation
drifts. The same update text resolves to different geometry, or to no geometry,
against a different cycle. The failure is silent without this rule.

## 3. A time query states which time axis it uses.

ADR-0036 has validity, ingress time, and observation time. It does not name the
two axes that a consumer selects between.

Add these terms:

- **Valid time.** What was in effect at time T. A restriction validity window.
  The cycle that was effective at T. A map time control uses valid time.
- **Knowledge time.** What the system held at time T. A snapshot handle ring, a
  replay of the ingress stream, and the immutable Briefing result use knowledge
  time.

State that one query uses one axis and reports which. Briefing already models
knowledge time. A map scrub needs valid time.

The reason for the rule: "the restriction was not yet active" and "we had not
yet received the restriction" are different facts. A single time control that
serves both reports one as the other.

## Acceptance

- [ ] ADR-0036 states that a tile bundle is a renderer encoding, not domain
      state, and that the bundle carries the cycle identity and the snapshot
      digest.
- [ ] ADR-0036 extends the coherence requirement to a rendered composition.
- [ ] ADR-0036 defines valid time and knowledge time, and requires a query to
      state its axis.
- [ ] The record stays `Proposed`. This issue does not accept it.

luofang34:
## Decision accepted

ADR-0036 is `Accepted`. ADR-0035 is `Superseded by ADR-0036`.
The acceptance commit changes both record statuses and both index entries.

Lifecycle is the ownership axis. Airmass owns weather state.
`AeronauticalUpdates` owns perishable notices and operational status.
Navdata owns the cycle-dated baseline. FlightPlanning owns plan state.
Briefing owns immutable evidence results from fixed inputs.

## Contract changes

- Valid time and knowledge time are separate query axes.
- A `SituationView` query fixes one axis and reports it.
- Briefing uses fixed revisions for knowledge state and a requested valid time.
- The Navdata snapshot identity contains the cycle, snapshot ID, and snapshot
  digest.
- The producer instance ID, subject identity, and revision define Navdata
  publication continuity.
- An application owns presentation, tiles, and styling.
- A rendered composition must use one navigation-data cycle.
- Link health is ownerless derived evidence.
- Migration step 1 marks the snapshot-envelope part complete.
- The diagrams now show `AirspaceView` and the fixed-input Briefing flow.
- The required *AirspaceView resolution contract: subject to baseline
  geometry* remains in #349.

## Review and validation

The proposed amendment received a separate review before acceptance.
The final review had no findings.

The local gates passed:

- `cargo fmt --all --check`
- `cargo test --all-targets`
- `cargo doc --no-deps` with missing-doc and broken-link denial
- `cargo build --release`
- structure, ADR lifecycle, and certification checks

Closes #348.
Refs #321 and #349.


luofang34:
## Verdict: accept after two required changes

The structure is right. Lifecycle is the correct axis for ownership, and the
record proves it in the alternatives section rather than asserting it. The
dependency rule ("a domain core must not depend on a source adapter or a source
protocol type") is stated once and then applied consistently, including the
worked FIS-B example. The two clock axes and the rule against comparing a raw
source clock with a monotonic clock are correct and are the kind of thing that
is very expensive to add later. Refusing to select a repository topology is the
right call and is defended.

Two findings block acceptance. Three should be fixed. Three are notes.

## Blocking

### B1. One evaluation time cannot express the query that Briefing already makes

The situation view section says "A query supplies an evaluation UTC value". One
value.

The Briefing paragraph says Briefing "reads fixed input revisions" and that its
input identifies "a plan revision, a requested time, a weather snapshot, an
aeronautical-update snapshot, and a navigation-data snapshot ID, digest, and
cycle".

Briefing therefore already separates two times: the state the system held (fixed
revisions) and the time the question is about (requested time). `SituationView`
cannot express that pair. Two components in one record model the same idea in
incompatible ways.

The same pair is needed outside Briefing. A time control on a map asks "what was
in effect then". A diagnosis asks "what did we hold then". These give different
answers, and a single evaluation value reports one as the other.

Required: name the two axes and state which one a query fixes.

- Valid time: what was in effect at the stated time. A restriction validity
  window. The navigation cycle that was effective then.
- Knowledge time: what the system held at the stated time. A snapshot handle, a
  replay position, an immutable Briefing result.

### B2. Navdata has no defined snapshot identity, and the record uses three names for it

The contract terms give every domain snapshot a producer instance ID and a
snapshot revision. The flight-plan handoff instead requires a "navigation-data
snapshot ID and digest, navigation-data cycle". The record never relates these
five identifiers.

A cycle-dated baseline does not fit the producer-instance model well. It changes
13 times each year on a published schedule, and every consumer keys on the
cycle. An implementer of Navdata must guess whether the cycle is the revision,
whether a rebuild of the same cycle is a new producer instance, and which
identifier a consumer compares.

Required: state how cycle, snapshot ID, digest, producer instance ID, and
revision relate for Navdata. An explicit exemption with its reason is also an
acceptable answer.

## Should fix

### S1. `AirspaceView` is named but not governed

Every other composed concept in this record gets rules. `SituationView` gets a
section. Briefing gets its input identified. `AirspaceView` gets one paragraph
that says what it combines and that geometry is optional.

It does not say which component resolves a subject identifier to baseline
geometry, which cycle that resolution uses, or what a failed resolution
produces. Those are not implementation details: they decide whether a notice
against a runway can be drawn at all.

The record already has a pattern for this. It defers the FIS-B exchange schema
to a "required record named *FIS-B assembly and classification contract*" and
defers execution policy to "*Component-edge delivery and supervision*". Use the
same pattern, or add the rules.

### S2. Link health has no owner

The table denies link health to `AeronauticalUpdates`. `aero-link` owns protocol
classification, not a lifecycle. The dependency section says link health "is not
authoritative `AeronauticalUpdates` state" and that "a domain can use link
health as quality evidence".

No component owns it. A product-unavailable report has a real lifecycle: a
product becomes unavailable, then available again. Either name an owner, or
state that link health is derived evidence that each consumer computes and that
no component stores.

### S3. Presentation has no owner, and the record denies maps four times

Surveillance, Airmass, `AeronauticalUpdates`, and Indicate each list maps in the
"does not own" column. A reader concludes that some component does own maps. No
row claims it.

One sentence fixes this: presentation, tiles, and styling belong to an
application; a tile bundle or a GeoJSON document is an encoding at a renderer
edge and is not domain state. This also settles where a cycle-dated baseline
goes when it is drawn, which is a live question.

## Notes

### N1. Migration step 1 is partly complete already

Step 1 is "Define the domain-owned ingress and immutable snapshot contracts".
The immutable snapshot half merged in #347 as `docs/domain-snapshot-envelope.md`
while this record was `Proposed`. The migration order reads as though all 11
steps are ahead. Mark the snapshot half complete; the ingress contracts are
still open.

### N2. The diagram omits `AirspaceView`, and one edge has no prose

`AirspaceView` is a defined concept and does not appear in the flowchart.

`BRIEF["Briefing"] --> VIEW` is the only edge with no explanation in the text.
Briefing is described as an immutable result computed from fixed inputs, so an
arrow into a live composition view needs a sentence, or the arrow points the
wrong way.

### N3. ADR-0035 status is correct as it stands

`main` has ADR-0035 at `Accepted`, not `Superseded`. That is right, because this
record says it supersedes "on acceptance". An earlier instruction in another
session asked for `Superseded` while this record stayed `Proposed`, which cannot
both be true. Leave ADR-0035 at `Accepted` and change it in the same change that
accepts this record.

## Disposition

#348 already carries B1 and S3, and #349 carries S1. B2, S2, N1, and N2 are new
and are added to #348.

Apply the changes to this branch while the record is `Proposed`. An amendment to
a proposed record is one edit. The same change after acceptance is a new record.

luofang34 (commented):
Re-review after the #348 amendment: no findings. The review checked the two time axes, Briefing input selection, Navdata snapshot identity and envelope scope, presentation ownership, rendered cycle coherence, link-health ownership, the #349 deferral, migration state, and both diagram corrections.

Issue #351, opened 2026-08-09
Give maplibre-rs an offline local tile source on iOS


luofang34:
Parent: luofang34/Pilotage#321

## The gap

`maplibre-rs` is the target renderer for the product client, because one Rust
codebase serves the browser client and the iPad client. It cannot serve an
aviation client today. The spike on #346 answered question 5 with No: there is no
way to read a tile bundle on iOS with no network.

The evidence, at upstream `main` `96b50a0`:

- The Apple entry point calls the non-web runner with no cache path. That runner
  constructs `ReqwestHttpClient`, which gets each tile with an HTTP request. Its
  optional disk store is an HTTP response cache, not a bundle reader.
- `StaticTileFetcher` embeds PBF files at build time. It does not read an
  installed tile directory, the renderer source client does not use it, and the
  Apple crate does not enable the feature.
- There is no MapLibre Tile decoder. The vector path decodes Mapbox Vector Tile
  data.
- Upstream issue maplibre/maplibre-rs#76 requests file, MBTiles, and PMTiles
  support. It is open.

## The seam

The core has a replaceable `HttpClient` trait in `maplibre/src/io/source_client.rs`.
An adapter that reads local data can implement it. The shipped iOS path neither
supplies such an adapter nor exposes an injection point to Swift.

So the work is: one local source adapter, plus a way for the iOS host to install
it.

## Work

- [ ] Implement a local tile source that reads an installed bundle. Choose one
      archive format and record the reason.
- [ ] Expose an injection point so the Apple host installs the source instead of
      the HTTP client.
- [ ] Show a map drawing from the bundle on a physical iPad with no network and
      with the radio interfaces disabled.
- [ ] Measure cold start and pan and zoom frame rate against the same scene on
      MapLibre Native.
- [ ] Decide whether the adapter goes upstream to maplibre/maplibre-rs#76 or
      stays local, and record the reason.

## Also carried here

These spike questions from #346 belong with this renderer, not with the MVP:

1. Does a raster layer draw a georeferenced grid at regional precipitation
   resolution? The MVP has no raster layer, so this gates version 2 weather
   imagery only.
6. Is there feature state, or another way to restyle an existing tile feature
   from live data?
7. Can two sources be held to one navigation-data cycle identity?

## Not a blocker for the MVP

The MVP uses MapLibre Native. The presentation adapter stays portable Rust, so
adopting `maplibre-rs` later changes the binding only.

luofang34:
[...]

Issue #350, opened 2026-08-09
Build a per-cycle baseline tile bundle from a Navdata snapshot


luofang34:
Parent: luofang34/Pilotage#321

## Objective

Convert a Navdata snapshot into a versioned tile bundle that a renderer draws.
Run the build once for each cycle, not for each frame and not on the device.

## Why this is a separate step

A National Airspace System Resources extract is large and changes every 28 days.
A perishable update is small and changes in seconds. They must not share a
delivery path. The tile bundle is an encoding at a renderer edge, like GeoJSON.
It is not domain state.

## Rules

- The build reads a Navdata snapshot. It does not read a source file directly.
- The bundle carries the cycle identity, the snapshot identity, and the snapshot
  digest.
- The bundle works with no network. An iPad in flight has no connection.
- The build runs in continuous integration or as an operator tool. It does not
  run inside a domain crate.
- A layer holds a stable subject identifier for each feature, so an overlay can
  name the same subject.

## Acceptance

- [ ] The build produces a bundle from a Navdata snapshot for one cycle.
- [ ] The bundle states its cycle identity and snapshot digest.
- [ ] A client reads the bundle with no network.
- [ ] Each drawable baseline feature carries a stable subject identifier.
- [ ] The build is reproducible: one snapshot gives one identical bundle.
- [ ] A size and build-time measurement is recorded for one full cycle.

## Open questions for the first step

- Which bundle format: PMTiles, an MapLibre Tile bundle, or a tile directory.
  #346 tests which of these the renderer reads.
- Which layers belong in the baseline bundle, and which stay live.

luofang34:
**Unblocked, and retargeted.** The blocker was the renderer, not the approach. #346 moves the minimum viable product to MapLibre Native, which reads MBTiles and an offline pack on iOS. A per-cycle baseline bundle is buildable again.

Target a MapLibre Native readable archive. Record which format and why. When `maplibre-rs` gains a local tile source, the same bundle should serve it, so prefer a format that both can read over one that only MapLibre Native reads.

The runway layer still depends on luofang34/v99n62#84, which adds the WGS84 runway-end positions that the Navdata model drops today.

Issue #346, opened 2026-08-09
Show traffic and weather in one maplibre-rs client, with a replaceable renderer binding


luofang34:
## Goal

Show traffic and weather from the domain components in one Pilotage client. Some
products must be visible at an earlier time, not only now.

## The renderer

`maplibre-rs` is a Rust map renderer on `wgpu`. It runs on macOS, Linux, Windows,
Android, iOS, and the web. A WebGPU backend for iOS and Android landed in
February 2026. It reads GeoJSON through `geojson-vt-rs`, and it has a `raster`
module.

`maplibre-rs` fits the Pilotage pattern better than a platform map SDK, because
one Rust codebase serves the wasm client and the iPad client. ADR-0032 already
uses that pattern: Pilotage keeps one portable runtime, and a thin platform layer
draws what the runtime produces.

`maplibre-rs` states its own limit: its goal "is not to replace existing vector
map renderers", and the implementation "serves as a proof-of-concept of the used
technology stack". Treat it as replaceable. The contract must not name it.

## The boundary

Three layers. Keep them separate.

1. **Domain snapshots.** `Surveillance` and `Airmass` emit typed, immutable,
   revisioned snapshots. See the domain snapshot envelope issue.
2. **A presentation adapter.** One portable Rust step converts a snapshot into
   renderer-neutral display values: a feature for each track, a grid for each
   gridded weather product, a shape for each advisory. This layer holds the
   display policy that no domain owns.
3. **The renderer binding.** `maplibre-rs` consumes the display values. A
   different renderer needs a different binding and no change above it.

A component must not send GeoJSON to another component. GeoJSON is an encoding at
layer 3.

## Across time

Two mechanisms. Do not join them.

- **A short scrub.** The composition layer keeps a bounded ring of immutable
  snapshot handles for each domain. A query at time T reads the handle for T.
  ADR-0036 permits this: the situation view "can keep a non-authoritative cache
  of immutable snapshot handles".
- **A long history.** Record the ingress stream and replay it through the same
  decoders. `Airmass` roadmap step 7 and `Surveillance` issue #11 already plan
  this. A recording reproduces any snapshot and costs no live memory.

ADR-0036 states the rule that keeps them apart: "Do not make a display object the
only copy of domain state."

## First work: one spike, before any client work

Answer the questions that decide whether `maplibre-rs` can serve this client:

1. Does a raster layer show a georeferenced grid at the resolution of a regional
   precipitation product? This is the least proven part and the one that weather
[...]

Issue #341, opened 2026-08-09
ARCH-03: define the read-only SituationView contract and corpus


luofang34:
Parent: luofang34/Pilotage#321

Depends on the acceptance of ADR-0036 in luofang34/Pilotage#330.

## Objective

Define the versioned, read-only `SituationView` request and result contract.
Add a conformance corpus in the same pull request.

## Contract rules

- The caller supplies evaluation UTC and freshness or coherence requirements.
- The composition host attaches its local monotonic evaluation stamp and clock
  identity.
- The view captures one immutable snapshot handle from each selected domain.
- The first contract uses best-available consistency. It does not claim an
  atomic update across domains.
- Each domain result carries a producer instance ID and snapshot revision.
- Each field or contributor carries source identity, source epoch, ingress
  time, source time, validity, quality, uncertainty, and a missing-data reason.
- The result reports ingress age and observation age separately. It reports an
  age as unknown when the required time evidence is not valid.
- Domain cores and the view contract do not depend on provider or link-service
  types.

## Acceptance

- [ ] Versioned request and result types define all required fields.
- [ ] The contract defines clock correspondence and its valid interval.
- [ ] The contract states its consistency guarantee.
- [ ] A deterministic corpus covers mixed-age inputs, unknown age, missing
  domains, source restarts, and clock-correspondence uncertainty.
- [ ] CI runs the corpus against each implementation.

PR #352, opened 2026-08-09
Build the iPad situation display with a replaceable binding

luofang34:
Tracks #346.

Summary:

- Package the pinned Surveillance and Airmass crates as one UniFFI XCFramework and Swift package.
- Add a portable Rust presentation adapter that owns traffic and weather display policy.
- Consume the typed `surveillance-geojson` lifecycle. Preserve `Upsert`, `Stale`, `Remove`, and `transfer_to`.
- Add a thin MapLibre Native 6.28.0 Swift binding for traffic points, weather points, and advisory polygons.
- Embed a pinned AeroLink driver copy with a separate host and driver App ID pair.
- Stage an app-owned USB wildcard entitlement that matches the self-service DriverKit development profile.
- Add boundary, dependency, entitlement, and coexistence guards with self-tests.
- Mark weather payload decoding and flight-category derivation as a stopgap. #353 tracks its removal after the Airmass feature boundary is available.

Local verification:

- Cargo format, Clippy, all-target tests, strict documentation, release build, and structure checks pass.
- Ten presentation tests and four FFI tests pass.
- The XCFramework builds for iOS, iOS Simulator, and macOS.
- The Swift core test and two GeoJSON edge tests pass.
- The arm64 iOS MapLibre binding target builds.
- ShellCheck and all new guard checks pass.
- Project generation and the deterministic two-host driver check pass.
- The complete Apple client pipeline builds the final arm64 simulator application from the pinned AeroLink source.

Device verification:

- A clean signed build uses the Pilotage host and driver App ID pair with self-service development profiles.
- The signed app and embedded DriverKit extension install on the attached trusted iPad.
- The installed host app launches successfully.
- The iPadOS Settings driver selection is not confirmed. This pull request stays in draft until that user-controlled check is complete.

GitHub verification:

- Read-only deploy keys provide Airmass, Surveillance, and AeroLink access.
- All eight push and pull-request checks pass on commit `c9b7739d`.

#341 landed independently. This pull request does not change its contract.

This change does not modify #350 or #351.

---
- <kbd>&nbsp;2&nbsp;</kbd> #358
- <kbd>&nbsp;1&nbsp;</kbd> #352 👈


luofang34:
The packaging, the guard scripts with self-tests, and the driver-coexistence
check are all right, and `check-situation-presentation-boundary.sh` correctly
stops the Swift binding from reaching `airmass_core`, `surveillance_core`,
`TrackSnapshot`, `WeatherSnapshot`, or `WeatherProduct`. Layer 3 is protected.

Two findings, both about layer 2 reaching into layer 1. One is cheap to fix now.

## Blocking: the traffic path bypasses `surveillance-geojson`

`crates/pilotage-presentation/src/traffic.rs` imports `surveillance_core` and
builds a `PointFeature` itself. `surveillance-geojson` merged in
luofang34/Surveillance#25 for exactly this, 33 minutes before this branch.

The cost is concrete. Feature identity is:

```rust
id: format!("traffic-{}-{}", handle.producer_instance_id().get(), track.id.get())
```

A `TrackId` changes when the engine merges a correlated duplicate. The feature
identity therefore changes, and the map deletes one feature and creates another.
That is the failure `FeatureDelta::Remove { transfer_to }` and
`RemovalReason::Merged { into }` exist to prevent: a merged duplicate must move
on the map, not disappear and reappear somewhere else. `docs/map-features.md` in
Surveillance states the reason.

Consume `surveillance-geojson` and keep its `Upsert`, `Stale`, and `Remove`
vocabulary. The crate is `no_std + alloc` and its GeoJSON encoding is optional,
so the dependency costs nothing that this crate does not already accept.

## Not blocking, but label it: weather decodes in the wrong component

`crates/pilotage-presentation/src/weather.rs` defines a Pilotage-owned media
type, `application/vnd.pilotage.weather-observation+json;version=1`, parses
`station_id`, `latitude_deg`, `longitude_deg`, `ceiling_ft_agl`, and
`visibility_statute_miles` out of a product payload, and derives the flight
category.

Two problems. Parsing a payload into typed fields is decoding, and ADR-0036
assigns decoding to the domain. Deriving a flight category from ceiling and
visibility is domain meaning, not display policy -- choosing the colors for VFR,
MVFR, IFR, and LIFR is display policy, and that part belongs here. Airmass
issue #9 already states the flight category as its own deliverable.

A second wire format for weather also competes with the Airmass typed values
that will arrive.

This is acceptable **as a stopgap**, because Airmass has no typed weather values
yet. It is not acceptable as an unmarked design. Mark it: say in the module
documentation and in `clients/apple-situation/README.md` that this path is
temporary, and open an issue to move payload parsing and flight-category
derivation into Airmass #9 and #7 when they land.

## Structural note, not a change request

The accepted ADR-0036 diagram routes a map adapter through `SituationView`. This
change composes `surveillance-core` and `airmass-core` directly in the FFI
session. That is reasonable today, because #341 does not exist yet. It does mean
the first real consumer ships before the contract it should use.

#341 has started. When its contract lands, this application should query
`SituationView`, and the #341 conformance corpus should treat this application
as its first required consumer. Without that, #341 becomes a contract with no
users and two composition paths exist.

luofang34:
Addressed in `898e1bd9`.

- `pilotage-presentation` now consumes `surveillance-geojson` at its merged revision. It preserves `Upsert`, `Stale`, `Remove`, and `transfer_to` through the Rust adapter, UniFFI façade, and Swift binding.
- Tests cover active updates, stale state, stale-revision rejection, removal, and merge transfer.
- The self-tested boundary guard rejects a direct `surveillance-core` traffic-policy import and rejects a missing producer adapter.
- The module documentation and client README identify weather payload decoding and flight-category derivation as a stopgap. #353 tracks removal of that stopgap. Airmass #9 is complete; Airmass #7 remains the upstream feature boundary.
- I noted the #341 structure. This PR does not add a dependency on that in-progress contract.

luofang34:
Hardware follow-up for `c9b7739d`:

- The original staged driver used exact vendor and product values. iPadOS rejected that signed extension with `0xe8008015` because the self-service development profile carries one USB vendor wildcard.
- The ignored AeroLink copy now receives an app-owned development entitlement with `{ idVendor = "*" }`. The pinned AeroLink source and exact `Info.plist` radio match table stay unchanged.
- The structure guard rejects integer or product values in that development entitlement. Its self-test proves the rejection. The generated-project coexistence check compares the staged entitlement with the app-owned template.
- A clean signed device build passed. The app and embedded driver installed on the attached trusted iPad, and the host app launched.

Apple documents the development wildcard and exact signing match in [How to sign a DEXT](https://developer.apple.com/forums/thread/809202). The iPadOS Settings driver toggle remains a user-controlled check, so this pull request stays draft.

PR #372, opened 2026-08-10
Replay a recorded reception file through the client pipeline

luofang34:
Seeing traffic or weather reach a map needs a radio, a transmitter in range, and
an iPad. That makes the whole path untestable at a desk, and it hides which stage
loses a product.

Run a recorded reception file through the same stages the client uses:
Surveillance, Airmass, and the presentation adapter. The counts say how far each
product travelled, and the layer breakdown says what a map would draw.

A weather report names its station and carries no position, so a run with no
navigation data draws no weather. The synthesize flag places every station it saw
at an invented point, which separates a weather path that cannot decode from one
that only lacks positions. The coordinates are not surveyed and the flag says so.

---
- <kbd>&nbsp;2&nbsp;</kbd> #373
- <kbd>&nbsp;1&nbsp;</kbd> #372 👈


luofang34:
Superseded: #367 already merged this file, including the helper split. Only the weather wiring remains, and that is #373 against main.

PR #376, opened 2026-08-10
Raise traffic and weather to the altitude they report

### LINKED ISSUE #374: Draw traffic at its reported altitude: a draped symbol hides the vertical picture
Traffic reaches the map as a circle or a text mark on a ground-plane source. With the
terrain renderer active, MapLibre drapes a symbol and a circle onto the elevation surface.
A target at 8,000 ft and a target at 800 ft therefore draw at the same place, and the
altitude survives only in the label.

The situation display is meant to answer "is that traffic above me or below me" at a
glance. A draped symbol cannot answer it, so the altitude that
`Carry the ADS-B altitude into the traffic display value` now delivers stops at the text.

## What the renderer supports

The pinned terrain renderer carries `fill-extrusion` and it composes with `terrain`: the
upstream branch regenerates a combined terrain and fill-extrusion baseline. A style has no
way to place a point symbol at a height, so an extruded polygon is the available path:

- one small polygon for each target, with `fill-extrusion-base` and
  `fill-extrusion-height` set from the reported altitude, draws a pad that floats at the
  reported height;
- adding a second polygon from the terrain surface to that pad draws a leader that shows
  which ground position the target sits above.

Both read from a GeoJSON source the presentation adapter already produces.

## Scope

1. `DisplayShapeStyle` gains the extrusion terms it needs: a base, a height, and whether
   the height is a ground offset or a mean-sea-level height. A style with no extrusion
   term keeps drawing as a fill and a line, so weather and airspace are unaffected.
2. The presentation adapter emits a traffic target as a shape with those terms when the
   track carries an altitude, and keeps the present circle when it does not. A track with
   no altitude must stay visible and must not claim a height.
3. The MapLibre binding adds `MLNFillExtrusionStyleLayer` for a shape style that declares
   extrusion.
4. The vertical datum has to be stated. ADS-B reports pressure altitude and geometric
   altitude, and the renderer's height is metres above the terrain surface or above the
   ellipsoid. A conversion that is left implicit will put a target at the wrong height over
   high ground.

## Guardrail

A test that feeds one track with an altitude and one without, and asserts the first
produces an extruded shape whose base and height match the reported altitude while the
second produces a point. The check must fail if a track with an altitude silently returns
to a draped circle.

## Depends on

Deciding how the terrain renderer is pinned, in
"Decide how to reach 3D terrain: the released renderer has none". `fill-extrusion` over
`terrain` needs that renderer.

luofang34:
A symbol and a circle are draped onto the terrain surface, so a target at 8,000 ft and a
target at 800 ft draw in the same place and the altitude survives only in the label. The
display is meant to answer "above me or below me" at a glance, and a draped symbol cannot.

The style spec has no way to place a point symbol at a height, so an extruded polygon is
the path. A shape style declares whether the renderer raises it, and a shape feature
carries the floor and ceiling it is raised between. A track with an altitude now also
produces a small pad at that height, above the symbol that keeps marking its ground
position. A track with no altitude keeps its symbol alone: an invented height would claim
knowledge the track never reported.

An advisory volume rides the same mechanism, with the floor and ceiling from the altitude
band Airmass decodes.

The height is measured from the terrain surface, which is what the renderer draws from.
A reported altitude is measured from mean sea level, so a feature over high ground rides
the terrain elevation too high. This crate cannot read the elevation model, and the label
carries the reported altitude and stays exact. Closes the display half; the datum
correction stays open.

Replay of the recorded five-minute reception produces twelve shapes where the flat build
produced four: eight traffic pads and four advisory volumes.

Closes #374.

---
This is **part 2 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #379
- <kbd>&nbsp;3&nbsp;</kbd> #377
- <kbd>&nbsp;2&nbsp;</kbd> #376 👈 

### DIFF
--- a/clients/apple-situation/Packages/PilotageGeoJSONEdge/Sources/PilotageGeoJSONEdge/FeatureCollection.swift
+++ b/clients/apple-situation/Packages/PilotageGeoJSONEdge/Sources/PilotageGeoJSONEdge/FeatureCollection.swift
@@ -47,12 +47,24 @@ public struct GeoJSONPolygonFeature: Equatable, Sendable {
     public let rings: [[GeoJSONPosition]]
     /// Ready-to-display text.
     public let label: String?
+    /// Floor of the shape, in metres above the terrain surface beneath it.
+    public let baseAboveTerrainMetres: Double?
+    /// Ceiling of the shape, in metres above the terrain surface beneath it.
+    public let topAboveTerrainMetres: Double?
 
     /// Create one polygon feature.
-    public init(id: String, rings: [[GeoJSONPosition]], label: String?) {
+    public init(
+        id: String,
+        rings: [[GeoJSONPosition]],
+        label: String?,
+        baseAboveTerrainMetres: Double? = nil,
+        topAboveTerrainMetres: Double? = nil
+    ) {
         self.id = id
         self.rings = rings
         self.label = label
+        self.baseAboveTerrainMetres = baseAboveTerrainMetres
+        self.topAboveTerrainMetres = topAboveTerrainMetres
     }
 }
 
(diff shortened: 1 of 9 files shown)


luofang34:

PR #377, opened 2026-08-10
Report what the client draws without a photograph of the screen

luofang34:
Answering "does the map show traffic at altitude and weather as areas" needed a person in
front of the iPad with a camera. A photograph captures whatever else is on the display, it
cannot be compared between runs, and it cannot say why a layer is empty.

The client writes the counts behind its map to Documents/situation-evidence.json, and a
script fetches the file over the same wireless connection that installs the build. The
file states the terrain build and archive, the state and byte and event counters of each
receiver, the feature count of each layer, the shapes the renderer raises and the heights
they span, and the tracks that reported no position.

The last field is the one that separates a broken pipeline from a quiet sky. A first run
against a live 1090 receiver reported 12.4 GB of completed samples, 17 accepted events, 0
rejected inputs, 3 tracked aircraft and 3 of them positionless. Nothing was wrong with the
client: no aircraft in range had transmitted a position.

---
This is **part 3 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #379
- <kbd>&nbsp;3&nbsp;</kbd> #377 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #376

### DIFF
--- a/clients/apple-situation/App/SituationClientModel.swift
+++ b/clients/apple-situation/App/SituationClientModel.swift
@@ -14,6 +14,7 @@ final class SituationClientModel: ObservableObject {
     private let domain: RadioDomainSession?
     private let terrainAvailable: Bool
     private let discovery = AeroLinkDiscoveryGate()
+    private let evidenceWriter = SituationEvidenceWriter()
     private var runtime: AeroLinkRadioRuntime?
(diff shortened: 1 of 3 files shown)


luofang34:

PR #379, opened 2026-08-10
Move recorded flights and reception off the map face

luofang34:
A radio, a transmitter in range, and weather in range all have to agree before the map can
be seen carrying traffic and weather. Off the device a recorded reception already answers
what the decoders and the display policy do with real products. It says nothing about the
renderer on the iPad.

A recording placed in the application container as
`Documents/replay-receptions-<instant>Z.ndjson` now drives the display through the same
decoders, the same display policy, and the same renderer. Nothing else starts it. While it
runs the client states that the map is a replay and not live reception, because a map fed
from a file looks exactly like a map fed from a radio.

Three things had to be right for a recording to reach the map, and each was silent when
wrong:

- The replay decodes through a domain session of its own. The live session counts
  reconnect generations for the radio it is attached to, and records tagged with a
  generation that session never issued are discarded as belonging to a replaced connection.
- Elapsed time starts from the device clock. The display policy never moves time backwards,
  so a replay based at zero leaves every replayed feature far enough in the past that the
  first live emission retires all of it.
- Wall-clock time stays the recording's. An advisory states its validity in real time, and
  a replay clocked from the device places every advisory outside its validity.

The evidence file now reports the recording, the records each stage produced, and the
records the display refused, so a replay that reaches nothing says which stage stopped it.

One run of the five-minute recording on an iPad Pro 11-inch: 15,191 receptions, 29,344
track records, 69 weather records, 0 refused, and a display holding one traffic pad
between 3,368 m and 3,428 m beside four advisory volumes.

---
This is **part 4 of 5 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #379 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #377
- <kbd>&nbsp;2&nbsp;</kbd> #376

### DIFF
--- a/clients/apple-situation/App/FlightsLibrary.swift
+++ b/clients/apple-situation/App/FlightsLibrary.swift
@@ -1,0 +1,100 @@
+import Foundation
+
+/// One recorded flight the client can replay.
+///
+/// A flight is a folder of what was observed during one run. Today it holds the reception
+/// file the harness writes, which already carries nearby traffic and every FIS-B product
+/// the receiver heard. An ownship track and the navigation-data cycle that was current
+/// belong to the same flight and are not recorded yet, so the type names the recording
+/// rather than the reception file: adding those later does not change what a flight is.
+struct Flight: Identifiable, Equatable {
+    /// Stable identity, the recording's file name.
+    var id: String { receptionFileName }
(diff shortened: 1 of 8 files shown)


luofang34:

PR #413, opened 2026-08-12
Dress the map modes panel in the platform's own chrome, and adapt it to narrow windows

luofang34:


luofang34:
Supersedes #412, which GitHub closed when its base branch was deleted on merge of #411. Same commits, rebased onto main.

PR #421, opened 2026-08-15
Point the source-identity comments at the decision in force

luofang34:
## What changes

ADR-0026 is superseded by ADR-0037. Two comments still cite the superseded decision,
which sends a reader to a document that no longer states the rule the code follows.


luofang34:
Closed by a branch delete that raced a failed merge. Continued as #423, same commit.

PR #520, opened 2026-08-25
Give the situation map a camera a pointer can reach

luofang34:
Stacked on #516.

## The problem

Two fingers already tilt and turn the map — the renderer's touch handlers are on by default, and the browser test proves a two-finger drag changes the pitch. A pointer has neither gesture, and the map carried no control at all, so a reader on a trackpad could not tilt or turn the map, and could not undo a tilt they reached another way. The map opens at pitch 55, so every reader starts tilted with no way back.

## The change

The map carries the renderer's compass, which turns the map when dragged and faces north when clicked, and a level control that looks straight down again. The level control appears only while the map is tilted — the rule the Apple client applies to both of its controls: a control that does nothing teaches a reader to ignore the corner it sits in.

`clients/web/situation-camera.js` holds the thresholds, the compass wording, and the move duration. Nothing in the map module restates them.

## Guardrails

- `situation-camera.test.mjs` reads `SituationCamera.swift`, `MapControlsView.swift`, and `SituationMapView.swift`: the 0.5° thresholds, the eight compass points and their spoken names, the two control sentences, and the 0.22 s move duration all fail here when one client changes without the other. It also pins that the two renderers' bearing conventions (GL JS normalizes to (-180, 180], Native reports 0 to 360) read as the same camera.
- The browser test drives the real controls in a real Chromium: it levels a tilted map, turns it with a ctrl-drag, tilts it with two fingers, and faces north again from the compass.
- `check-web-situation-map.sh` refuses a map without both controls, and refuses camera thresholds or wording restated in the map module.


luofang34:
Independent review found five items on this PR; all are fixed in a second commit.

**MEDIUM — the compass already levels the map, so the level control was a second answer to the same question.** With `visualizePitch: true`, `NavigationControl`'s compass handler calls `resetNorthPitch`, which eases bearing *and pitch* to zero. The added control did half of that at 220 ms against the compass's 1000 ms — the exact divergence the shared vocabulary exists to prevent — and the module comment claiming "nothing in it looks straight down again" was contradicted by the `visualizePitch: true` two lines below it.

**MEDIUM — the level control overlapped the compass and stole its clicks.** Measured from `maplibre-gl.css`: three 29 px buttons with a 10 px margin put the compass at y 70–99; the control at `top: 5.9rem` spanned 94.4–123.2 with coincident right edges, and it was appended after `addControl`, so it won the z-index tie over the bottom 4.6 px of a 29 px target.

**MEDIUM — it outlived the map.** `Map.remove()` removes the renderer's own DOM and leaves anything the page added, so on the load-deadline path a visible button remained, its `move` listener was never removed, and a click called `easeTo` on a destroyed map. Stripping the `maplibregl-map` class also removed the positioned ancestor it was anchored to.

All three are answered by removing the control. The compass is the whole answer.

**LOW — `isRotated` was computed and never used, and the parity claim was not true.** MapLibre's compass is always visible, so "a control appears only when there is something to undo" held on Apple and not here. The comments and the document now say why instead: touch reaches the camera without a control, so the Apple client can hide one; a pointer reaches the camera only through the compass, so the compass stays. `isRotated` now decides what the map *says* about itself rather than being unused.

**LOW — the accessible name was rewritten every animation frame.** `move` fires per frame during a drag, and an accessible name rewritten at that rate is announced at that rate. It is written only when the heading changes.

Also corrected: the module said GL JS normalizes to `(-180, 180]`; it is `[-180, 180)`. The reviewer confirmed the two clients' thresholds agree over every bearing either renderer can report, and that a value outside one full turn is unreachable from both — the comment now says that rather than claiming a stronger equivalence.

One bug of my own that the browser test caught while fixing this: removing the level control took `loadStylesheet` with it, and the map reported `MAP_RENDER_FAILED` with the real `ReferenceError`. The typed-reason contract did its job; the test now prints the state and reason in the failure message so the next such break names itself.

Browser test now drives: ctrl-drag turns (0° → 292.8°) and tilts (55° → 60°), the compass faces north and looks straight down (292.8° → 0°, 60° → 0°), two fingers moving together tilt from level (0° → 57.5°), and the map's own label reads "Facing north" when it does and "Facing north west, turn back to north" when turned.

luofang34:

Every commit on this branch (`e2784d21`) is reachable from `main` — verified with `git merge-base --is-ancestor`. The work landed through the squash of the lane above it, so there is nothing left for GitHub to merge here and merging would produce an empty or misleading commit.

PR #521, opened 2026-08-25
Make the coastline archive complete over the world

luofang34:
Stacked on #516 and #520.

**This replaces the earlier "world floor" approach in this PR.** The floor
was a second, shallower archive drawn under the detailed one. Review found
that its coarse water painted over detailed land — 14.8 km² in one window,
up to 150 m inland — so hydrography was drawn from the detailed source
only, and that left the flat rectangle of background over the Atlantic that
started this rework. Two sources at two resolutions could not both be
correct at the seam. There is now one source and no seam.

## What was wrong

A raster tile the terrain archive does not hold is drawn from the tile
above it, stretched. A vector tile the coastline archive does not hold
draws **nothing**, and no shallower tile stands in for it. The archive held
a global band only to zoom 5 and deeper bands only over one region, so past
zoom 5 outside that region the land and the sea stopped being drawn.

## What this does

The archive is one band and that band covers the world, zoom 0 through 7.
Above zoom 7 the renderer stretches the tiles it has, so the picture
changes with the zoom and never with where the reader is looking.

Depth is what completeness costs: zoom 7 over the world is 21845 tiles and
about 13 MB; each deeper zoom multiplies both by four; and Natural Earth
1:10m carries no more shape than zoom 7 draws.

## Guard

The invariant is now stated rather than approximated: every zoom the plan
declares covers the world, exactly one band covers each zoom, and the
closest zoom is the deepest band. When an archive is present the guard also
reads it, and fails if the deepest zoom holds fewer tiles than global
coverage produces.

The self-test gained a `mutate` helper that fails when a mutation changes
nothing. That is not hypothetical: a stale `closest_zoom: 15` pattern had
stopped matching the plan, so the guard read an untouched fixture, accepted
it correctly, and the case reported a hole that was not there.

## Measured

Archive: 21845 tiles, 13.3 MB, full 4^z count at every zoom 0–7.
Exported TileJSON declares `maxzoom: 7`, which is what makes the renderer
overzoom instead of requesting a tile that does not exist.

Frames captured with `gl.readPixels`, counting pixels still showing the
style's background colour:

| view | zoom | bare background |
|---|---|---|
| Atlantic, off New Jersey | 7 | 0.00% |
| Atlantic, off New Jersey | 10 | 0.00% |
| Zurich | 12 | 0.00% |
| mid-Pacific | 9 | 0.00% |
| Indian Ocean | 4 | 0.00% |

Every one of those lay outside the old regional and local bands, and three
are deeper than the archive's deepest tile.

### DIFF
--- a/clients/apple/Resources/SituationCoastline.manifest.json
+++ b/clients/apple/Resources/SituationCoastline.manifest.json
@@ -1,12 +1,12 @@
 {
-  "plan_sha256": "1ea2b7719ae2fce191ce3160d5c6424c5aef0abcf65777275481406dabd5a095",
-  "archive_sha256": "d54e07e7be017d822470ddf0f2587361047b77d2ced1ed54e6c4e520a5fda89e",
-  "tiles_written": 17171,
-  "archive_bytes": 7995392,
+  "plan_sha256": "ba97418e145471e7aac00e217a6b22e2d4c8be6a3d3fda902f82ca480f77d69c",
+  "archive_sha256": "9cf9cbefca1fea731b873888da82110207492f920677f7811190d104a3432df7",
+  "tiles_written": 21845,
+  "archive_bytes": 13955072,
   "source_name": "Natural Earth",
   "dataset_scale": "1:10m",
   "attribution": "Made with Natural Earth. Free vector and raster map data at naturalearthdata.com.",
-  "closest_zoom": 15,
+  "closest_zoom": 7,
   "sources": [
     {
       "name": "ocean",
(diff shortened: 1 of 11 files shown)


luofang34:
Independent review found one item on this PR, and confirmed the mechanism.

**HIGH — the floor's declared zoom ceiling was never asserted.** `UPDATE metadata SET value = … WHERE name = 'maxzoom'` succeeds silently when it matches no row, and that row is the entire mechanism: it is what MapLibre Native reads from the `mbtiles://` archive to decide the source's zoom range. A floor whose metadata said 15 would hold only z≤5 tiles, pass every check, and produce exactly the blank coastline this PR exists to prevent. The guard now asserts the row.

The reviewer also verified what this PR claims:
- the extraction is **byte-reproducible** — the same SQL run twice hashed identically and matched the committed `world_archive_sha256`;
- `minzoom`, `bounds`, `scheme`, and `format` are already correct in the copied metadata, so only `maxzoom` needed the update;
- "a source takes its zoom range from the archive it reads" is **true for MapLibre Native** (the style spec gives no way to override `maxzoom` on a `url` source, and a layer-level `maxzoom` cannot make a source overzoom) and **not how the web path works** — the web resolver writes an explicit `tiles`/`minzoom`/`maxzoom`, so on the web a second source over the same tile template would have sufficed. The second archive is necessary for Apple and redundant on the web; the document now says that rather than presenting the constraint as universal.

**Known gap, not fixed here:** CI never runs `build-situation-coastline.sh`, so every check inside `if [ -f "$archive" ]` — the digest match, the tile-depth check, the new metadata check, and the pre-existing `vector_layers` check — runs only on a machine that has already built. That is pre-existing (the terrain guard had the same shape), and closing it means adding GDAL to the CI image. Filed rather than folded in here.

luofang34:

Every commit on this branch (`aa260e43`) is reachable from `main` — verified with `git merge-base --is-ancestor`. The work landed through the squash of the lane above it, so there is nothing left for GitHub to merge here and merging would produce an empty or misleading commit.

PR #522, opened 2026-08-25
Give the base map an aeronautical chart look

luofang34:
Stacked on #516, #520, #521. Rebased after #521 was reworked from a world
floor to a globally complete archive; the floor-specific commit in this PR
became obsolete and is gone.

## Palette

The elevation tints follow the FAA VFR sectional ladder, with the
sectional's own band edges at 1000, 2000, 3000, 5000, 7000, 9000, and
12 000 ft converted to metres. The values come from the legend of the FAA
Aeronautical Chart User's Guide, because the FAA publishes no colour table;
two editions agree to within a few steps of 255. Above the sectional's
highest band and below sea level the ladder is ours, and the docs say so.

Hillshade uses the Igor method — it darkens a slope without staining the
tint under it — lit from the north west as a chart is.

## Water reads as water

Water draws **over** the relief and the hillshade. An elevation ramp cannot
tell a lake at 500 m from ground at 500 m, so the polygon that knows keeps
its own colour. Open water and inland water read apart.

## Drainage

Three files feed the river layer. The global 1:10m network ranks 0–9 and
draws from the zoom its rank earns. The regional files for Europe and North
America carry the minor drainage a chart is read by and rank 10–12, so the
ladder holds them back until zoom 9. The lakes layer takes the two regional
files too.

**Stated limit:** the rivers are centre lines and stay centre lines.
Natural Earth publishes no river areas, so a river has width only where a
lake polygon gives it one. A sectional chart draws rivers the same way and
thins them by rank; this style follows that, and widens a line with the
zoom rather than pretending to a shape the data does not carry. At zoom 13
a centre line visibly disagrees with the valley the relief shows — the
source is generalised at 1:10,000,000. The provenance and the docs both say
so rather than leaving a reader to discover it.

## Builder

Each source now declares the geometry it carries and the fields it keeps.
GDAL writes a line source into a polygon layer with only a warning, and
that layer then draws nothing. Fields are chosen with `-sql` rather than
`-select`, because GDAL refuses `-select` on an append and several sources
now feed one layer.

## Guards

New: hydrography draws over the relief; open and inland water read apart;
river width thins by the rank the data carries; the plan's layer set is
exactly ocean/land/lakes/rivers however many sources feed them; the built
archive carries all four vector layers.

The self-test gained five style mutations and a check that a style mutation
which changes nothing fails the case instead of reading as a guard hole.

## Measured

Archive: 21845 tiles, 19.8 MB, vector layers `["ocean","land","lakes","rivers"]`.
A zoom-7 tile over Zurich carries rivers at ranks 2, 4, 8, 10, 11 — the
regional supplement reaches the tiles.

Frames captured with `gl.readPixels`, bare-background share:

| view | zoom | bare background |
|---|---|---|
| Zurich | 9 | 0.00% |
| Zurich | 12 | 0.00% |
| Rhine at Waldshut | 11 | 0.00% |
| Aare | 13 | 0.00% |
| Atlantic | 10 | 0.00% |
| Europe | 3 | 0.00% |

### DIFF
--- a/clients/apple/Resources/SituationCoastline.manifest.json
+++ b/clients/apple/Resources/SituationCoastline.manifest.json
@@ -1,8 +1,8 @@
 {
-  "plan_sha256": "ba97418e145471e7aac00e217a6b22e2d4c8be6a3d3fda902f82ca480f77d69c",
-  "archive_sha256": "9cf9cbefca1fea731b873888da82110207492f920677f7811190d104a3432df7",
+  "plan_sha256": "b550edbe89de100b4f296dc17e054aa19510ac038394a972b66916a5b3490170",
+  "archive_sha256": "0f0b52d3bf0109052b0fca9b8690853b9c66c88e6ead9fd027596678dd42eaec",
   "tiles_written": 21845,
-  "archive_bytes": 13955072,
+  "archive_bytes": 20758528,
   "source_name": "Natural Earth",
   "dataset_scale": "1:10m",
   "attribution": "Made with Natural Earth. Free vector and raster map data at naturalearthdata.com.",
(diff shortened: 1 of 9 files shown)


luofang34:
Independent review found a real regression I introduced here, plus guard holes that let it through. Both are fixed in a third commit.

**HIGH — the world floor's water painted over the detailed land.** The floor is one zoom band stretched at every deeper zoom. Its water was opaque and drawn above the relief, and no detailed land layer sits above that, so wherever the coarse ocean polygon disagreed with the detailed one it repainted real ground as open water and nothing restored it. The reviewer measured it against the archives with GDAL: **14.8 km²** of coast in one window (max 150 m inland), **63.3 km²** in a Chesapeake window across 468 pieces.

No ordering fixes this: land must sit under the relief to take the elevation tint, so no order puts coarse water above the relief and detailed land above that water. The hydrography now comes from the detailed archive alone, which is exact wherever it draws. Where that archive does not reach, water takes the terrain tint — the map draws no shoreline it cannot place, and never draws one over ground. Documented as the tradeoff it is.

**HIGH — the guards pinned four unpaintable layers and left every visible one unpinned.** Because the opaque water fills covered identical geometry from the same source-layer, the four `*-fill` layers the guard pinned colours on were dead paint, while the tones a reader actually sees were unguarded. The reviewer proved ten regressions passed the guard green, including recolouring the visible ocean to magenta and switching the whole world floor off.

The guard now pins the visible water and river layers by type, source, source-layer, tone, and order against the *hillshade*; requires the two water tones to read apart; requires the river width to thin by the rank the data carries; forbids hydrography from the world floor; and compares floor against detail on everything that decides whether a layer draws — `layout`, `filter`, `minzoom`, `maxzoom` — not only colour.

**The five assertions this PR added had no self-test.** That is how the holes survived. Nine fail-before halves added, one per assertion.

Worth recording: two of my first drafts of these assertions were themselves vacuous — a jq recursive descent inside `all()` evaluates against the array element, not the document, and `all()` over an empty stream is true. The self-test caught both, which is the argument for writing it.

**LOW — `scalerank` was carried into every tile and read by nothing**, so the whole world's 1:10m drainage drew as a web of hairlines at 0.3 px (4312 features at z2). River width now thins by rank.

**MEDIUM — the background was still near-black** under a chart-white map, so the map flashed black then white. It is now neither the land tone nor the water tone: an unpainted map must not read as either.

**Documentation** corrected throughout: the token count (three → the tokens the style carries), the relief/hillshade naming collision, the provenance definition that said "polygons" while the archive carries lines, the guard message that still said "ocean, land, and lakes", and the FAA provenance overreach — the ramp continues above the sectional's highest band and below sea level where the legend stops, and that part is ours, not the FAA's. The below-sea-level tint reuse (it shares the 1000–2000 ft tone, as the sectional does) is now stated rather than left for a reader to discover.

Verified after the fixes: guard, guard self-test (12 mutations), style test, camera test, structure test, and both browser gates pass; a real-archive capture confirms the chart look with no coarse water over land.

luofang34:

Every commit on this branch (`ac51e1a7`) is reachable from `main` — verified with `git merge-base --is-ancestor`. The work landed through the squash of the lane above it, so there is nothing left for GitHub to merge here and merging would produce an empty or misleading commit.

PR #523, opened 2026-08-25
Carry a geodetic position on the wire

luofang34:
## Why

The wire carries only local north-east-down metres. Nothing on it says where that local frame sits on the Earth, so no client can place a vehicle on a map. The iPad works around this with CoreLocation — which answers where the *reader* is, never where the aircraft is.

## What this adds

A `GeodeticFix` message: latitude, longitude, height, and every datum identity the height needs. `crates/pilotage-geo` already owns the typed contract that refuses a datum a reader cannot interpret, and the value crosses through it at both ends, so an unknown datum, a missing realization, or an MSL height that names no separation is refused rather than displayed.

**An absent fix is absent.** There is no default anywhere on the path. The existing wire idiom substitutes `0.0` for an absent group and relies on a companion stamp to mark it meaningless — safe for a NED metre, fatal for a coordinate: 0,0 is Null Island, a real place in the Gulf of Guinea, and a receiver would draw a plausible vehicle there. The fix is a sub-message that is simply absent instead.

**Estimate and oracle stay apart**, as the rest of the sample does:
- the estimator's fix rides `AvionicsState` with its own stamp, because a GNSS solution advances independently of the local-frame group;
- the oracle's fix rides `SimTruthState` under that sample's stamp, because it comes from the same observation, and the truth lane's existing role gate refuses it in a lane stamped as an estimate.

Reusing one message type in two differently-stamped places preserves that separation for free; one shared field would have destroyed it.

Every producer emits `None` here. This PR is the contract, not a lane — the first lane is the next PR.

`pilotage-geo` also gains `SIMULATOR_GEOID_MODEL_ID`, which lets a simulated MSL height have a legal vertical datum without pretending to name a geoid. Named-constant precedent: `SIMULATOR_PROFILE_ID`, `SIMULATOR_CONFORMAL_POLICY_ID`.

## Field numbering

`AvionicsState` 22/23, `SimTruthState` 13 — all previously free. `buf` runs `breaking: FILE`, so the file placement is permanent; the message lives in `telemetry.proto` beside its carriers.

## Guardrails

- Host test: an absent fix stays absent, asserted explicitly against Null Island.
- Host round-trip test: the value a receiver rebuilds through `GeodeticPosition::new` equals the value sent.
- `clients/web/geodetic-fix.test.mjs`: twelve uninterpretable datums are each refused (unknown datum, unknown code, NAD83 without a realization, MSL without a geoid, AGL without terrain, baro without a setting, local-relative without an origin, latitude past the pole, longitude at and past the wrap point), and a fix inside a mislabeled truth lane never arrives.
- `scripts/check-geodetic-datum-codes.sh` + self-test: the Rust enum, the proto comment a reader implements from, and the browser decoder stay in step; a datum must reach the wire through `to_u8()` rather than a cast; an absent fix must not take a default.

## Verification

`cargo fmt`, `cargo clippy --all-targets` (clean), `cargo test --all-targets` (green), `check-structure`, and both new guards pass locally.


luofang34:
Independent review found ten items; every one is addressed. The branch was rebuilt from `origin/main` so the two commits carry source only.

**Structural (finding 9): 21,785 build-artifact files were swept into the commit.** The branch was cut from `origin/main`, whose `.gitignore` does not yet carry the vendor and situation-asset entries added on the map lane, so `git add -A` took a 7.7 MB archive and a whole exported tile tree with it. Rebuilt with explicit paths: 22 source files, no artifacts.

**HIGH — the estimate lane had no role gate (finding 2).** Every other lane gates on its own role; the new field gated on stamp presence only, so an oracle position placed on `AvionicsState.geodetic` reached the client as the estimator's own GNSS solution. The browser now gates it on the estimate role, and the browser test drives six wrong roles. The proto comment claiming a physical profile publishes no truth was removed — `mavlink_batch` publishes unconditionally, so the sentence was false.

**HIGH — the u64 origin decoded through a 53-bit path (finding 3).** Two origins differing in the lowest bit compared equal past 2^53, which is precisely what the origin identity exists to make visible. It decodes through `firstBigVarint` now; the test pins two origins that previously collapsed.

**MEDIUM-HIGH — accuracy 0 read as perfect (finding 4).** `health_from_accuracy` rates 0 mm as the best possible fix, and proto3 omits a zero, so silence and a perfection claim were the same bytes. Zero is now documented and decoded as UNSTATED (`null`), never a number a reader can derive health from.

**MEDIUM — the guard was defeated by `cargo fmt` (finding 5).** The line-local `map_or` pattern stopped matching once the formatter wrapped the expression, so the guard caught only the form CI's own fmt gate forbids. It now checks for what a default needs (`GeodeticFix::default`) and for the exact mapping expression, neither of which wraps.

**MEDIUM — three JS/Rust asymmetries (finding 6).** A datum sent under the wrong wire type decoded to `NaN` and passed every range test; `realization` and `geoid_model` are `uint32` on the wire but `u16` in Rust, so a value past `u16` was displayed after truncating to something the contract refuses; and `geodetic_to_wire` never validated, so a producer using a struct literal could put +180 or NaN on the wire. All three are closed: every identity must be a whole number inside its Rust type's range, and the mapping re-runs `position.validate()` before sending.

**MEDIUM — `GeoidModelId(1)` was unreserved (finding 7)** and already used as an arbitrary round-trip value elsewhere in the crate. It is `u16::MAX` now, at the top of the range a registry numbering from the bottom cannot reach, and the mapping refuses it outside the simulation-truth role — the doc-only rule is now structural, with a test.

**Test gaps (finding 8):** the `AvionicsState` path had zero coverage, which is why finding 2 shipped; the Null-Island assertion was decorative (`notDeepEqual(null, {...})` can never fail). Both fixed, plus cases for an empty-but-present message, an all-zero message, an origin past 2^53, an identity past its type, and an unstated accuracy.

Findings 1 and 10 belong to the lane and are answered in #524.

Verified after the fixes: `cargo fmt --check`, `cargo clippy --all-targets` clean, `cargo test --all-targets` green, `check-structure` OK, both guards and the guard self-test pass, all 8 browser decoder checks pass.

PR #524, opened 2026-08-25
Publish the simulator oracle's position on the Earth

luofang34:
Stacked on #523.

## What was wrong

The link already decoded latitude, longitude, and altitude from every
HIL_STATE_QUATERNION report. It projected them into the local frame against
an origin latched from the first report, then dropped the report itself, so
the position the simulator stated could not reach any client.

## What this does

The report travels beside the projection, as a typed fix under the same
stamp, because the fix and the local frame are one observation.

**A report that states no position is refused whole**, not only its
geodetic half. The local frame is that position projected, so a projection
of no position is a position too: measured against a latched origin an
all-zero report published a point 5266 km from the vehicle, under a flag
that says the position is valid, beside a geodetic half that correctly
refused it. The refusal now runs at the front, where both halves are still
one observation.

**The projection subtracts in f64.** Two wire i32 values differ by more
than an i32 holds: a debug build panicked at the ends of the range and a
release build wrapped into a garbage frame. A vehicle crossing the
antimeridian reaches those ends, and a test now holds it.

**A restart releases the latched origin.** Every other cached group was
cleared on a new acquisition epoch and this one was not, so a simulator that
restarted into another world measured the new world's positions from the old
world's datum — the geodetic fix saying where the vehicle was and the local
frame saying where it was not, in the same sample.

**No length gate.** An earlier revision required 48 payload bytes so a
truncated report could not invent a position. Review showed that wrong both
ways: MAVLink 2 trims trailing zero BYTES and can cut into the middle of a
value, so a vehicle at rest at 500 m sends a 47-byte frame that the gate
dropped entirely — costing the attitude and the velocity too — while a
full-length report at a zero longitude still latched the origin at
Greenwich. What a zero means is settled per field instead, and a test holds
the trimmed frame.

## This is the X-Plane lane

X-Plane publishes no shared-memory block, so no truth oracle attaches and
the forwarded frame stays authoritative — true before this PR and asserted
nowhere. Two tests now chain wire bytes → link cache → published typed fix,
each proven to fail when broken:

- `a_real_truth_frame_carries_its_position_into_the_cache` fails when
  latitude is read from longitude's offset. The velocity assertion the codec
  suite already carried does not move under that mutation.
- `the_x_plane_lane_publishes_the_position_the_flight_controller_forwarded`
  fails when an absent oracle's sample is made authoritative.

Offsets and CRC_EXTRA were verified against the reference implementation
(pymavlink), not read off a documentation page: id 115, CRC_EXTRA 4,
lat/lon/alt at 36/40/44 as degE7/degE7/mm.

## Simulation only

`sampling.rs` compiles in a flight build but only reads what the link
decoded; nothing here adds a simulation dependency.
`scripts/check-flight-build.sh` runs in CI at `ci.yml:162`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1711,6 +1711,7 @@ dependencies = [
  "pilotage-adapter-api",
  "pilotage-control-feel",
  "pilotage-control-web",
+ "pilotage-geo",
  "pilotage-input",
  "pilotage-mavlink",
  "pilotage-protocol",
--- a/adapters/aviate/Cargo.toml
+++ b/adapters/aviate/Cargo.toml
@@ -14,6 +14,7 @@ aviate-xil-shm = { workspace = true, optional = true }
 getrandom = { version = "0.3.4", features = ["std"] }
 pilotage-adapter-api = { workspace = true }
 pilotage-control-feel = { workspace = true }
+pilotage-geo = { workspace = true }
 pilotage-mavlink = { workspace = true }
 pilotage-sim-video = { workspace = true, optional = true }
 pilotage-protocol = { workspace = true }
--- a/adapters/aviate/src/adapter/sampling.rs
+++ b/adapters/aviate/src/adapter/sampling.rs
@@ -3,15 +3,19 @@
 use std::sync::{Arc, Mutex};
 
 use pilotage_adapter_api::{
-    AvionicsAttitudeSample, AvionicsKinematicsSample, AvionicsSample, Pose2d, SourceRole,
-    TelemetryBatch, TelemetrySample,
+    AvionicsAttitudeSample, AvionicsKinematicsSample, AvionicsSample, GeodeticFixSample, Pose2d,
+    SourceRole, TelemetryBatch, TelemetrySample,
 };
 use pilotage_protocol::VehicleId;
 use pilotage_timing::SimTick;
 
 use super::WITHHOLD_AFTER;
+use pilotage_geo::{
+    BaroSettingId, DatumRealizationId, GeodeticPosition, HorizontalDatum, LocalOriginId,
+    PositionQuality, SIMULATOR_GEOID_MODEL_ID, TerrainRefId, VerticalDatum, VerticalPosition,
+};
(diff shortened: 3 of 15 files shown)


luofang34:
Independent review found two items on this lane; both are fixed.

**HIGH (finding 1) — a zeroed truth report minted a valid Null Island fix.** The absent-fix guarantee in #523 covers absence, not a report that states zeros. A `HIL_STATE_QUATERNION` with `lat = lon = alt = 0` produced a fully valid fix at 0,0 that nothing downstream could refuse: 0,0 is a real place off the coast of Africa, so the typed contract accepts it by design and cannot be the filter.

This is reachable, not hypothetical. `decode_sim_truth` reads every field through zero-extending accessors, so a short or zero-trimmed payload is indistinguishable from a genuine zero report, and a simulator that streams truth before its position initialises produces the same bytes.

The lane refuses the all-zero triple at the one place that knows what it means. The reviewer also caught the aggravating half: the same report latched `truth_origin` at 0,0,0 for the life of the link, poisoning the NED projection too. A report that states no position no longer becomes the origin, with a test in `pilotage-mavlink` that pins both halves — no origin latched, no truth published, and the first stated position latching correctly afterwards.

**LOW (finding 10) — `pub(crate)` was not the minimal change.** `sim_truth_sample` was widened so a test in a different module tree could call it, which is unconditional in release builds for a test-only need. The tests moved to `adapters/aviate/src/adapter/sampling/tests.rs`, the sibling module the repo convention and the `shm_sampling` precedent both point at, and the visibility is back to private.

The branch was also rebuilt from the cleaned #523 tip; see the note there about the build artifacts that had been swept in.

Verified after the fixes: `cargo fmt --check`, `cargo clippy --all-targets` clean, `cargo test --all-targets` green, `check-structure` OK. The lane's own suite is 13 tests in `sampling/tests.rs` plus the origin-latch test in the link crate.

luofang34:

Every commit on this branch (`9d6a6131`) is reachable from `main` — verified with `git merge-base --is-ancestor`. The work landed through the squash of the lane above it, so there is nothing left for GitHub to merge here and merging would produce an empty or misleading commit.

PR #540, opened 2026-08-27
Start each vehicle on its own control-feel law

luofang34:
## What shipped

`hosts/session-host/src/runtime/aviate_profile.rs` picks the control-feel artifact with no vehicle in the decision:

```rust
None => ValidatedFlightFeelProfile::from_json_str(ALIA250_DEFAULT_CONTROL_FEEL_JSON)
```

Both backends hand the host `--adapter aviate`, so it cannot tell the vehicles apart. Absent an override every session loaded that one compiled-in default — the Alia's law. **The X500 in gazebo started on it**, confirmed in its own logs:

```
aviate-gz    (x500): feel_profile_id=alia250-legacy-v1
aviate-xplane(Alia): feel_profile_id=alia250-legacy-v1
```

Three shaped X500 laws were already in `adapters/aviate/profiles/` with nothing to select them, and the override the host reads already existed. Only the wiring was missing.

## Why the launcher

Independently confirmed during review: the host hardcodes `HOST_VEHICLE: VehicleId::new(1)` for every adapter, and nothing in the Aviate adapter's production code, the link, or the FC handshake carries an airframe identity the host could switch on. Deriving it host-side needs new Aviate-side plumbing. The launcher is the only component that knows which aircraft it started.

## Scope — this binds the START law only

A runtime feel-mode request does **not** reach these files. `request_feel_mode_under` builds `FlightFeelProfile::shaped(mode)` — not `shaped_for(airframe, mode)`, which production code never calls — and `shaped` hardcodes `alia250-shaped-{mode}-v1` with unfitted numbers.

So the X500 starts on its fitted law and moves off it the moment an operator touches the feel control. My own verification session shows exactly that:

```
2  alia250-shaped-agile-v1
2  alia250-shaped-balanced-v1
2  alia250-shaped-precision-v1
```

This is a pre-existing hole in the adapter, not introduced here, and this PR does not fix it — `request_feel_mode` cannot be fixed from the launcher. Stated so the binding is not read as more durable than it is. Follow-up: carry airframe identity to the adapter so `shaped_for` is reachable.

## Corrections to an earlier version of this PR

An earlier body claimed this fixed `pose Missing` / `velocity Missing` / `MeasurementUnavailable`. **That claim was wrong and has been removed.**

The decisive fact: `pose Missing` and `velocity Missing` are literals in the BROWSER — `clients/web/telemetry-display.js:43` and `:49` — and appear zero times in any Rust source. The table quoted the web client rendering a null field, which is in no position to know which artifact the host loaded.

Host-side, those fields come from `planar_projection`, gated on attitude and kinematics freshness within `WITHHOLD_AFTER`, an estimator-status observation having been published, skew coherence, and the FC's velocity valid bit — every input a field of `LinkState`, fed by MAVLink from the FC. The control-feel profile is installed on the UPLINK, the outbound command path. It never writes `LinkState` and never participates in sampling.

The third row was the same cause again: `apply_flight` returns `MeasurementUnavailable` on a missing pose BEFORE reaching `answer_feel_requests`, so all three rows were one condition printed three times.

Both runs also loaded a valid artifact through the identical validator, and a rejected one would have exited the host at startup — no session at all, rather than a session with degraded telemetry.

Worst of all, the PR contradicted itself: it argued the feel layer shapes command into setpoint and so cannot explain divergence with nothing commanded, then four paragraphs earlier claimed it explained missing measurements. Measurement availability is upstream of the feel layer and on the opposite direction of flow.

Likely confounds between the two sessions: estimator warm-up or FC bring-up timing, a `source_epoch` mismatch after a sim reset, or simply a different Aviate binary — #538 and #539 landed immediately before this branch.

The provenance argument stands on its own and is the whole justification.

## Two defects this PR introduced, both fixed here

1. **`--profile physical` regression** (`d5a3429e`). The host does not ignore a named law there — it refuses the session with `AviatePhysicalControlFeelOverride`. `aviate-gz` is the only backend without a `plan()` profile gate, so that was a live working path that would have become a startup failure. `aviate-xplane` was never exposed.

2. **Clobbering the operator's own law** (`a4eafa61`). Stage env is applied on top of the inherited environment, so writing the key replaced an operator's `PILOTAGE_AVIATE_CONTROL_FEEL_PROFILE` with no diagnostic anywhere — the same "law nobody chose" failure, pointed the other way, and the path a tuning session uses. Now defers to an ambient value, as `RUST_LOG` and `PILOTAGE_XPLANE_AIRFRAME` already do.

## Guardrails — all three verified fail-before

- `each_aviate_backend_names_its_own_vehicles_control_feel` — each backend names a law that exists and belongs to the aircraft it launches. Asserts **exactly one** entry: env applies in order and the last write wins, so a correct entry followed by a stale one launches the stale law. Review demonstrated that mutation surviving the first version; it now fails with `names the control-feel law 2 times`.
- `a_physical_session_is_handed_no_control_feel_law`
- `an_operators_own_control_feel_law_is_not_replaced` — and asserts the no-ambient case still binds, so it cannot be satisfied by never naming a law.

The host-side loader test now covers all six shipped laws, not just the three Alia ones.

## Known residual

The env key is a bare string in two crates that share no dependency, so renaming it host-side leaves the launcher naming nothing while the tests stay green. Noted in a comment at the constant; closing it properly wants a shared crate.

Also: naming the Alia's file converts a compiled-in guarantee into a runtime file read, so a missing file now yields `AviateControlFeelRead` where it previously could not fail. The Alia stays on the compatibility law — which belongs to no airframe — deliberately, because moving it to `alia250-shaped-*` changes what the aircraft does and wants a flight behind it.

## Checks

`fmt --check` clean, `clippy -p xtask --all-targets -D warnings` zero, `rustdoc -D broken_intra_doc_links` clean, 73/73 xtask tests, host profile tests pass.

### DIFF
--- a/hosts/session-host/src/runtime/aviate_profile.rs
+++ b/hosts/session-host/src/runtime/aviate_profile.rs
@@ -113,6 +113,21 @@ mod tests {
                 "alia250-shaped-agile-v1.json",
                 pilotage_control_feel::FeelMode::Agile,
             ),
+            // The X500's, which the launcher names by path for a gazebo
+            // session. This host is what reads that path, so a file it
+            // cannot load is a session that does not start.
+            (
+                "x500-shaped-precision-v1.json",
+                pilotage_control_feel::FeelMode::Precision,
+            ),
+            (
+                "x500-shaped-balanced-v1.json",
+                pilotage_control_feel::FeelMode::Balanced,
+            ),
+            (
+                "x500-shaped-agile-v1.json",
+                pilotage_control_feel::FeelMode::Agile,
+            ),
         ] {
             let path = profiles.join(file);
             let loaded = control_feel_from_env_blocking(
--- a/tools/xtask/src/backend.rs
+++ b/tools/xtask/src/backend.rs
@@ -137,6 +137,66 @@ mod tests {
     use super::backend_for;
     use crate::error::XtaskError;
 
+    /// Every Aviate backend must name a control-feel law belonging to the
+    /// aircraft it actually launches.
+    ///
+    /// The host cannot check this for itself: both backends hand it
+    /// `--adapter aviate`, so absent an explicit profile it loads one
+    /// compiled-in default for every vehicle. That default is the Alia's,
+    /// and the X500 flew on it — a law qualified for another airframe.
+    /// The launcher is the only component that knows which aircraft it
+    /// started, so this is where the binding is asserted.
+    #[test]
+    fn each_aviate_backend_names_its_own_vehicles_control_feel() {
+        use super::SessionContext;
+        use crate::cli::Profile;
+        use std::path::{Path, PathBuf};
+
+        let repo_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .parent()
+            .and_then(Path::parent)
+            .expect("workspace root is two levels above tools/xtask")
+            .to_path_buf();
+        let ctx = SessionContext {
+            repo_root: repo_root.clone(),
+            host_port: 4433,
+            viewer_port: 8080,
+            profile: Profile::Simulation,
+            log_dir: repo_root.join("target/xtask-sim"),
+            lan: false,
+        };
+
+        // Backend name -> the substring its profile's file name must carry.
+        for (backend, vehicle) in [("aviate-gz", "x500"), ("aviate-xplane", "alia250")] {
+            let env = backend_for(backend).expect("known backend").host_env(&ctx);
+            let (_, value) = env
+                .iter()
+                .find(|(key, _)| key == "PILOTAGE_AVIATE_CONTROL_FEEL_PROFILE")
+                .unwrap_or_else(|| {
+                    panic!(
+                        "{backend} names no control-feel law, so the host falls back to \
+                         whatever single default it was compiled with"
+                    )
+                });
+
+            let path = PathBuf::from(value);
+            let name = path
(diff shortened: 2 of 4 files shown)


luofang34:
## Self-review found a regression this PR introduced — fixed in `d5a3429e`

The host does not *ignore* a named control-feel law on a physical profile. It refuses the session:

```rust
Some(path) if profile == AviateProfile::Physical => {
    Err(HostError::AviatePhysicalControlFeelOverride { path })
}
```

The first commit named a law unconditionally, so `cargo xtask sim --fc aviate-gz --profile physical` would have gone from working (fell through to the default) to a startup failure. `aviate-xplane` was never exposed — it already rejects any profile but `simulation` — but `aviate-gz` was.

The decision now lives in one `control_feel_env` helper both backends call, so a backend added later cannot reintroduce it by writing the variable itself, and the refusal is stated where the decision is made rather than rediscovered from the host's error type.

Second guardrail, `a_physical_session_is_handed_no_control_feel_law`, asserts the empty case directly. Without the guard:

```
aviate-gz names a control-feel law for a physical session, which the host
refuses with AviatePhysicalControlFeelOverride
```

Both guardrails now have a verified fail-before half.

Gates: `fmt --check` clean, `clippy -p xtask --all-targets -D warnings` zero, `rustdoc -D broken_intra_doc_links` clean, **72/72** xtask tests.

PR #541, opened 2026-08-27
Make the launcher the only writer of the control-feel variable

luofang34:
Follow-up to #540, closing the one mutation that survived its review.

## The hole

A backend can still clobber an operator's own control-feel law and keep every test green — hand-roll the push, guard `Physical` by hand, forget the ambient value:

```rust
if ctx.profile != Profile::Physical {
    env.push((
        "PILOTAGE_AVIATE_CONTROL_FEEL_PROFILE".to_owned(),
        ctx.repo_root.join(CONTROL_FEEL_PROFILE).display().to_string(),
    ));
}
```

73/73 passed with that in place.

Neither test could see it. `a_physical_session_is_handed_no_control_feel_law` goes through `host_env`, so it catches a bypassing backend for the physical case. `an_operators_own_control_feel_law_is_not_replaced` calls the helper directly, so a backend that never calls the helper is invisible to it.

That left the helper's doc promising more than the tests delivered — it claimed "a backend added later cannot reintroduce either by writing the variable itself", when only the physical half was enforced at the seam.

## The fix is structural, not another assertion

Naming a vehicle's law and deciding whether to pass it are separate jobs, and only the second is subtle — it has two cases where passing does harm rather than nothing.

A backend now DECLARES its law:

```rust
fn control_feel_profile(&self) -> Option<&'static str> { Some(CONTROL_FEEL_PROFILE) }
```

and `run_session` applies it. A backend has nothing to remember, and no way to write the variable even deliberately.

## Guardrail

`no_backend_writes_the_control_feel_variable_itself` asserts it directly, across every backend × every profile. On the bypass above:

```
aviate-gz writes PILOTAGE_AVIATE_CONTROL_FEEL_PROFILE in host_env for Simulation;
it must declare the law through control_feel_profile() and let the launcher
decide whether passing it is safe
```

That is the mutation that survived #540. The other three tests move to the same seam: the binding is read off the declaration, and the physical case is asked of the decision rather than of one backend's assembled environment.

Also picks up two review nits: a past-tense doc fragment and a stray empty `///`.

## Checks

`fmt --check` clean, `clippy -p xtask --all-targets -D warnings` zero, `rustdoc -D broken_intra_doc_links` zero, 74/74 xtask tests.


luofang34:
## M-J and M-K both fixed — `1776338e`

### M-J: the binding was verified nowhere the launcher assembles it

Correct finding, and it was this PR's doing. I first tried the extracted-function-plus-direct-assertion you suggested, and **your mutation still survived it** — the test proves `host_env_for` is right, not that `run_sim` uses its result. Keeping the call alive and pushing `backend.host_env(&ctx)` passes a direct test of the join.

So the fix is to remove the parameter rather than assert about it. `host_stage` no longer accepts an environment:

```rust
pub(crate) fn host_stage(ctx: &SessionContext, backend: &dyn SimBackend) -> Stage
```

It derives one. There is nothing left to pass wrongly at the call site, and the test asserts on the **Stage that is pushed** — the last artifact before a process is spawned, with nothing between it and the host.

The nearest surviving equivalent (make `host_stage` build env from the backend alone) now fails:

```
the environment aviate-gz hands the host names the control-feel law 0 times;
the host reads one value and the last write wins
```

The exactly-one assertion is back with it, for the reason you gave.

### M-K: "no backend" over a hand-written list

Taken, in the stronger form you proposed. `BACKENDS` is a table and `backend_for` resolves by iterating it, so a backend absent from the table cannot be selected at all — the failing direction is the safe one. The contract tests iterate the same table.

I registered your `jsbsim-gz` rogue backend verbatim:

```
jsbsim-gz writes PILOTAGE_AVIATE_CONTROL_FEEL_PROFILE in host_env for Simulation;
it must declare the law through control_feel_profile() and let the launcher
decide whether passing it is safe
```

Caught automatically, with no test edit.

### Also

`_ctx` nit taken.

Six mutations across this lane, six kills: M-E, M-F, M-H, M-G, M-J, M-K.

`fmt --check` clean, `clippy -p xtask --all-targets -D warnings` zero, `rustdoc -D broken_intra_doc_links` zero, 75/75 xtask tests.

luofang34:
Final two commits after the review verdict: `af1079bb` splits the test module to `backend/tests.rs` unchanged (the structure gate refuses source files past 500 lines — the registry tests pushed backend.rs to 565); `75b8afa2` adds the registry name-agreement invariant, which independently caught the reviewer's M-L mutation verbatim ("the table registers alia250-xplane but the backend it builds calls itself aviate-gz") and was verified to fail on a deliberate disagreement. Content otherwise exactly as reviewed: six mutations, six kills.

PR #554, opened 2026-08-28
Run every guard pair through one discovering runner

luofang34:
## Stage D of the agreed CI redesign

- **`cargo xtask guards`**: discovers every `scripts/check-<name>.sh` with a `scripts/test-check-<name>.sh` sibling (19 pairs today), runs each self-test before its check, and continues past failures so one red run names every failing guard. Discovery is the contract: **adding a guard never edits CI again.**
- **ci.yml**: 37 steps replaced by one (19 pairs; one self-test — `test-check-flight-build.sh` — existed on disk but was never wired into the workflow, so the runner *adds* that coverage).
- **Policy in AGENTS.md** (STE): shell starts programs; decisions live in Rust with tests; migrate-on-touch for old guards, deleting the shell self-test in the same change — per the agreed design, the debt stops growing now and drains at the rate guards actually change.
- Tests: a synthetic-directory discovery test (pairing semantics) and a real-repository test pinning that the known guards — including the Linux-run apple-client static guard — are discovered.

## Verification

`cargo xtask guards` locally: **19 pairs ok** (full output in the run log). clippy `-D warnings` clean, fmt clean, check-structure OK, xtask tests 79/79, workflow parses, zero paired invocations remain in ci.yml.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,29 +128,10 @@ jobs:
       - name: cargo clippy
         run: cargo clippy --all-targets -- -D warnings
 
-      - name: Production Rust lint guard self-test
-        run: sh scripts/test-check-production-rust-lints.sh
-
-      - name: Production Rust lint guard
-        run: sh scripts/check-production-rust-lints.sh
-
-      - name: Monotonic counter arithmetic guard self-test
-        run: bash scripts/test-check-monotonic-counter-arithmetic.sh
-
-      - name: Monotonic counter arithmetic guard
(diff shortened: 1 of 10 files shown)


luofang34:
Amendment pushed: the first full local run passed because ci.yml still carried the literal steps; after the folding, two guards' own CI-wiring assertions (`CI must run scripts/...`) correctly went red — they assert CI names them. Second commit generalizes exactly that assertion (paired guards satisfied by the `-- guards` runner step; unpaired obligations still literal) and the full run is again 19/19 with exit checked this time. Reviewer: treat those two scripts' assertion hunks as first-class review targets.

PR #599, opened 2026-09-17
feat(apple): add the pinned Indicate HWD review bench (blocked on an unpublished Indicate rev)

luofang34:
## What this is

This draft publishes the HWD review bench for the iPad app as it stands on the workstation: `HwdReviewView`, a menu entry, and the build script for the bench.

## Blocked on Indicate

The commit pins Indicate revision `0e268cfc0826cc81ca9b4025299ff44fa178aa62`. That revision is not on the Indicate remote. It is in a local Indicate lane that has uncommitted work and an open agent session, so it was not pushed from here. CI cannot fetch the pin until that lane is published.

## Verification

Rust gates pass on the stack head (see #598). The Swift change is not verified here.

Stacked on #598.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #603, opened 2026-09-17
chore(indicate): advance the Indicate pin to state ABI v8 (owner decisions inside)

luofang34:
## What this does

Advances all eight `indicate-*` pins from `985eb33` to `9b2c2b8` (head of Indicate `main`, the `v0.6.3` tree plus an Indicate lockfile change) and migrates Pilotage from state ABI v7 to v8.

- Rust consumers import `abi::v8`. The feed declares no airspeed trend, because no lane measures one.
- The browser state writer writes version 8, the three longer groups (Air 16, Nav 43, Dynamics 20 bytes), Trust bit 9, and the four new groups in a new module. It matches Indicate's three v8 golden frames byte for byte.
- The identity pins that move with the revision are updated: `clients/instrument-compatibility.json`, the runtime corpus constants, the browser gate, the Apple bridge tests, the Swift gate constants, and the evidence-gate workflow revision.
- One guardrail was repaired: the digest-mismatch tests changed the first hex digit to `0`, and the new corpus digest starts with `0`, so that case had become vacuous.

## Review and the decisions that remain

An independent review (2026-09-17) read the 28-file diff against the v8 codec at `9b2c2b8` and compared every browser encoder offset, width and default byte with the Rust encoder. It found no defect in the migration: no v7 offset survives, the goldens are read from the pinned revision and not hand-edited, and every pin agrees (`9b2c2b8` / ABI 8 / corpus 8 / `0b0c7ccb…`). It found that decision 3 below rested on a false premise, and it gave the exact change that makes the Apple job green.

1. **The Apple client followed corpus 8.** `IndicateAppleDisplay` `77304c95` (luofang34/IndicateAppleDisplay#12) reports corpus 8 with the digest that this branch pins; the consumer package pins that revision now. The two new corpus cases decode and paint with the interpreter as it is.
2. **Tag or head.** The earlier pin was a tagged revision. `v0.6.3` (`3bfeac3`) is the same tree for Pilotage.
3. **The stored panel index is dead state.** `HostLinkModel` persists `selectedPanel`, but no code reads it: each tile resolves through `panelChoice(forTileId:)` by descriptor id. A stored `2` selected nothing before and selects nothing after. A follow-up can delete `selectedPanel` and `panelKey`.
4. **No Pilotage source feeds true airspeed, the airspeed trend, or the four new groups.** They stay absent. Both live clients omit the whole air group, so the Missing state of the true-airspeed box is observable on the bench only.
5. ADR-0031 does not yet say that the nav scale is declared as Terminal.

## Verification (on a second machine)

Pass: `cargo check --workspace --all-targets`; tests of the runtime, feed, web shell, web and Apple bridge crates (52 tests); `cargo clippy --workspace --all-targets -- -D warnings`; `cargo fmt --check`; `check-indicate-compatibility.sh`; `check-structure.sh`; `check-production-rust-lints.sh`; docs with `-D missing_docs`; the nested FFI workspace (`--locked` clippy, 54 tests); `build-web-instruments.sh`; the instrument node suites (state-abi, instruments with 198 checks, scene-conformance with 45 cases and 0 divergences, and 8 more).

**Not run there:** `swift test` for the Apple instrument consumer (that machine has no Xcode toolchain), the `*.browser.test.mjs` suites, and the node suites that are not about instruments. The CI job `apple-instrument-consumer` runs the Swift gate.

## Context

Indicate's local workstation lane is 20 commits past this revision and adds `IngressConfig.role`. That work is not published, so this PR does not depend on it. #599 pins one of those unpublished revisions and stays blocked until it is.


luofang34:
`apple-instrument-consumer` at `a1deaa6` (corpus-8 pin) passed every Swift test, including "The Swift gate equals the Pilotage consumer pin", and then died in the bridge integration probe: `ProducerFault(reason: .stateBadVersion)`. `GeneratedBridgeIntegration.swift` wrote a state frame with a typed version byte `[7, 0]`; the v8 runtime refuses it. The probe and the one test with the same literal now take the version byte from `AppleInstrumentCompatibilityGate.stateABI`, so the next ABI advance cannot leave them behind. Verified on claw: the probe steps of `scripts/check-apple-instrument-consumer.sh` (bridge build, generated bindings, IndicateAppleDisplay and consumer libraries, the probe executable) pass; `swift test` and the simulator build need Xcode and run in CI.

PR #604, opened 2026-09-17
feat(agent): agent core, model port and a model evaluation harness (ADR-0042)

luofang34:
Stacked on #602. This is the follow-up that #602 names: the move from a standalone tool to the client composition.

## What this adds

- **ADR-0042 (Proposed).** An agent is a client module with a shared core and platform ports (ADR-0037). It holds leased scopes as a principal (ADR-0025). A model is a separate process behind a model port. Pilotage links no model and selects no model.
- **`crates/pilotage-agent`**, the shared core, with no I/O and no wire types: the directive vocabulary (ATC-style: `direct_to`, `heading`, `altitude`, `speed`, `hold`, `join_procedure`, `land`, `go_around`, `return_to_base`, `takeoff`, `unable`), the model port with frames for one or more image sources including a panorama, three checks before a directive is flown, the executor, the verifier, and the scores for a model.
- **`tools/intent-pilot`** as the headless port: `intent-pilot` (scenario or live operator messages) and `agent-eval` (one score for every model, with an append-only run ledger).
- Five adapters behind the same port, and two suites (one open, one held-out).

## Measured

Held-out suite, 46 cases, first run of each adapter (the ledger records the run number):

| Adapter | Correct | Wrong and flown | Wrong and stopped |
| --- | --- | --- | --- |
| `gemma4:e4b` through Ollama | 45 | 1 | 0 |
| `Qwen2.5-1.5B-Instruct-4bit`, JSON decoding | 36 | 7 | 3 |
| keyword baseline, no model | 31 | 3 | 0 |
| parallel constrained decoding, the same Qwen weights | 19 | 5 | 19 |

The same person wrote the suite and the adapters, so this is not a blind test. The comparison between adapters is fair. Two leaks of held-out wording into the baseline and the shared prompt were found and removed before the first run.

Gazebo flights with `gemma4:e4b` on the Aviate x500, verdict from stamped simulator truth: `vectors` 5 of 5, `approach` 3 of 3, `go-around` 3 of 3, `direct-land` 2 of 2, `retarget-recall` 0 of 3 (the model reads a message with two instructions as the first one only; with the oracle adapter the scenario passes). The host rejected no control frame.

## Defects that the flights found, now corrected with tests

1. **A false pass.** `retarget-recall` had a checkpoint for its second message and none for its first. A flight that did not fly the first leg passed. The expectation now has an ordered checkpoint for each leg.
2. **A made-up destination.** For "Proceed direct ZULU and land." a model replied `ALPHA`. `ALPHA` is permitted, so the envelope check passed it, and the vehicle started to fly there. The grounding check now refuses a name, a number or a turn side that is not in the words of the message. In a second live run the same message was refused.
3. **An old armed report.** After a flight-controller restart the telemetry still said armed, and the executor sent a climb demand to a disarmed vehicle for 150 s. A climb that does not leave the ground in 3 s now sends a new arm request.
4. **A vehicle left in the air.** The pilot stopped at the verdict. It now brings the vehicle home first.

## Gates run

`cargo fmt --all --check`; `cargo clippy --all-targets -- -D warnings` on the two crates; 65 tests; docs with `-D missing_docs`; `cargo check --workspace`; release build; `check-structure.sh`; `check-production-rust-lints.sh`; `check-flight-build.sh`; `check-monotonic-counter-arithmetic.sh`; `check-adr-supersession.sh`. The binary of this commit flew `vectors` to a pass.

## Open, and named in the ADR

The mission core of ADR-0041 has no `heading`, `speed` or `go_around` action, and its document does not change in flight, so the agent core has an executor of its own for now. The host grant path does not mark an agent as different from a person. A message with two instructions needs a decision. No adapter reads frames yet; the port and the harness carry them.


luofang34:
## Independent review of the agent core, and what changed

A second reviewer read `crates/pilotage-agent` and ADR-0042 at the head of this stack, with the instruction to break the executor, the checks, the verifier and the harness. It found 1 blocker, 5 should-fix items and 7 nits. Each is answered below; the fixes are the four commits from `f27180a` on this branch, with tests that failed before each fix.

| # | Finding | Answer |
| --- | --- | --- |
| 1 | **Blocker.** `land`, `hold present position` and `takeoff` were dropped without a word in some phases, and the record showed a flown directive. | Fixed. `Refusal::NoEffect` refuses a directive that changes nothing, so the record shows the refusal. A landing during the climb descends at once; a hold during the climb holds at the present height; a hold, a landing or a return during arming cancels the takeoff and disarms; a return on the ground or during the disarm is refused (it is a takeoff that nobody asked for, and the headless port sent one on `quit`). |
| 2 | `altitude` during a descent did not stop the descent. | Fixed. A new height ends a descent where the vehicle is. |
| 3 | `on_arrival` was neither envelope-checked nor grounded; `land`/`return_to_base`/`takeoff`/`go_around` passed grounding with no word; a number after "heading" grounded a height. | Fixed. The arrival must be the one that was said; each slotless directive needs its word; a number after a heading or turn word, or before a degree word, binds to a heading. A test grounds the expected directive of every case of both suites in its own message, so the check stops no correct answer (it did not, in a third held-out run of each adapter). |
| 4 | The verifier read `armed: None` as disarmed. | Fixed. Once a vehicle has reported an arm state, a landing needs a disarmed report. |
| 5 | The range hold trapped every later heading. | Fixed. A heading that points back toward the launch point leaves the hold. |
| 6 | No crate entry point bound the three checks. | `AgentFlight::take_reply` in #607 binds them, and the headless pilot and the web port both fly through it. |
| 7 | `flying_seconds` did not measure "the newest directive". | The documents now state the real rule: the newest phase in the air, restarted by an arrival, a hold or a new directive. The go-around scenario relies on it. |
| 8 | Unit variants of the scenario accepted unknown fields; `ModelReply` and the suites lack `deny_unknown_fields`. | The scenario variants are struct variants now. The reply keeps extra members open: the port adds a request number, and an adapter may add diagnostics. |
| 9 | The executor clamps a speed to the advertised maximum. | Left. The envelope check refuses a speed above the maximum before the executor sees it; the clamp is a floor under a wrong envelope. |
| 10 | A commanded turn side yields to the short way under 20°. | Left, and documented in the guidance: a side that is 15° away is a short turn either way. |
| 11 | A scenario test was vacuous (renaming ALPHA also broke the procedure). | Fixed. The test renames a fix that nothing else names and asserts the reason. |
| 12 | ADR-0042 claimed a declaration field for state and history, a number-slot declaration, and a hazard reader that do not exist. | Fixed. The record states the port as it is. |
| 13 | A disarm request can go out at 1.4 m when the flight controller stops the descent early. | Left. Aviate refuses a disarm in the air and the executor asks again each 2 s; the README names this. |

The reviewer also confirmed: the guidance frames and signs, the arm and force-arm logic, the arrival gate, the verifier's inputs (truth, the arm report and the clock; no model output, no phase), the scoring (nothing scored against model output), the check order in the host, and the crate rules (no `unwrap`/`expect`/`panic` outside tests, sizes, `wrapping_add`).

Issue #605, opened 2026-09-17
MISSION: add heading, speed and go-around flight actions


luofang34:
## Problem

The flight action vocabulary of the mission core (ADR-0041) has `arm`, `climb`, `follow_plan`, `maintain_target`, `land` and `disarm`. An operator instruction in air traffic control phraseology also uses a heading ("turn left heading 270"), a speed ("reduce speed to 2") and a go-around. The agent client module of ADR-0042 lowers a directive onto a flight action. It has no flight action for these three directives, so it keeps an executor of its own.

## Scope

- Add `heading`, `speed` and `go_around` to `FlightAction`, with validation, the capability mapping and stable action names.
- Keep each host honest: a host that cannot execute a new action returns a typed refusal. It does not map the action onto a different action.

## Out of scope

- Execution of the new actions in the operational host (`pilotage-mission`).
- A change of the active target after a mission starts.

## Acceptance

- Each new action round-trips through canonical JSON.
- A heading outside 0 to less than 2π, a speed that is not greater than zero, and a value that is not finite are refused.
- The engine gives the action to the host without a change, and a go-around can be the cleanup action of an aborted landing.

PR #607, opened 2026-09-17
feat(agent): fly the agent as a client module of the browser viewer

luofang34:
Stacked on #604. Refs #605 for the mission-core actions that the executor will use later.

## Summary

The agent flew as a standalone tool with a connection of its own. This change makes it a part of the operator client, as ADR-0042 and ADR-0037 describe: it reads what the client reads, and it feeds control through the path that a person's input takes.

| Part | Where | What it does |
| --- | --- | --- |
| Automation input source | `clients/web-control` | `InputSource::Agent` beside the keyboard and a pad, in the control runtime that the web client and the Apple binding share |
| One flight type | `crates/pilotage-agent` | `AgentFlight` owns the model request, the three checks, the executor and the state. The headless pilot now flies through it too |
| Web port of the agent core | `clients/web-agent` | wasm surface: JSON text for the request, the reply and the decision; numbers for the control tick |
| Browser module | `clients/web/agent-*.js` | agent panel, model port, inputs from the client's own telemetry and advertisement |
| Model gateway | `tools/intent-pilot` (`model-gateway`) | the model port over HTTP for a client that cannot start a process |

## How the agent enters control

- An engage and a release take the transactional neutral handover of a device change. The lease stays held. The activation revision advances.
- The announcement names `automation.intent-pilot/v1` as the device profile while the agent is the source. The host log of the live run shows `Keyboard` → agent → `Keyboard` → agent.
- The demand does not go through a stick mapping. The runtime bounds each value, reads a value that is not finite as neutral, and flies neutral in a mode that has no velocity law.
- An arm request of the agent is the typed arm edge of the runtime, so the live-lease gates apply.
- **The first operator input wins.** A deflected control or a safety press, on the keyboard or the pad, releases the agent. The press that takes control does not also fire. The handover holds neutral until the operator's controls are neutral.
- The demand then uses the client's control loop, its single publish gate, its typed intent builder and its action tracker. No second lease encoder, no second gate.

The model stays outside Pilotage. The browser sends the request line of the shared core to the gateway and gives the reply line back to the core. The shell reads no directive. The gateway holds no authority, listens on loopback by default, and answers a page only when its origin is in the permitted list.

## Live check in a real browser

Headless Chrome on a second machine, the viewer and the gateway through `localhost` tunnels, WebTransport direct to the session host, `gemma4:e4b` behind the gateway. Each check reads the vehicle pose from Gazebo (`gz model -p`), not from the agent and not from the page.

| Run | Result | Notes |
| --- | --- | --- |
| 1 | 10 of 11 | The failed check was the first probe of my rig (it read the wrong element for the lease). Each flight check passed. |
| 2 | lost | My rig stopped on a failed truth sample. The browser closed, and the link-loss policy of the host landed the vehicle. |
| 3 | 11 of 11 | 5.25 m after takeoff; heading 264 and 10.3 m west after 7 s; 0.9 m from `BRAVO`; an operator key released the agent and the vehicle stayed at 5.1 m; after a second engage, "Cleared RNAV 27 approach." ended 0.19 m from `HOME` on the ground. |
| 4 | 11 of 11 | Heading 271 and 13.9 m west after 7 s; 0.7 m from `BRAVO`; landed 0.45 m from `HOME`. |
| 5 | 11 of 11 | Heading 267 and 13.5 m west after 7 s; 0.4 m from `BRAVO`; landed 0.30 m from `HOME`. This run is the recorded demonstration video. |
| 6 | 11 of 11 | Flown after the fix of the release source (below). Heading 265 and 10.3 m west after 7 s; 0.9 m from `BRAVO`; landed 0.28 m from `HOME`. |

In each of the six runs the model gave `ALPHA` for the fix `ZULU` that does not exist, and the grounding check refused it. The host rejected no control frame.

## A defect that a review found after the first flights

A release of the agent always made the keyboard the source. With a pad selected, the shell kept sending pad samples, so the announcement named the keyboard profile while the pad drove. That is the attribution hazard of CTRL-HAZ-01. The browser flights did not show it, because they took control with a key. The coordinator now remembers the operator source that the agent displaced, follows a pad connect or disconnect under the agent, and returns to that source. An engaged agent also keeps its identity: a second identity was accepted with no handover and no new revision, and it is now refused. Three Rust tests and one browser check on the real wasm failed before the fix (`"Keyboard"` where `"Sony DualSense"` was expected).

## Verification (Apple M4, 2026-09-17)

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, clippy for `wasm32-unknown-unknown` on the two wasm crates, `cargo doc` with `-D missing_docs -D rustdoc::broken_intra_doc_links`: pass.
- `cargo test --locked -p pilotage-agent -p pilotage-agent-web -p pilotage-control-web -p intent-pilot -p xtask`: 259 passed, 1 failed. The failure is `viewer_server_sets_cache_control_no_store` in `xtask`. It fails in the same way on a worktree of `main` on this machine (3 of 3 runs), so it is a fault of this environment and not of this change. I did not find its cause.
- `check-structure.sh`, `check-production-rust-lints.sh`, `check-monotonic-counter-arithmetic.sh`: pass.
- All 50 node tests of `clients/web` pass. The new `agent-module.test.mjs` runs 35 checks on the real agent wasm and the real control wasm, with a scripted model.
- Evidence gate (`indicate-evidence` at the pinned revision), `--require-resolvable`: only the pending review slot remains, as on `main`.
- Mutation checks, each caught by exactly one test: no stick override; no safety-press override; no flight mode guard; no bound on the demand; a pad selection that takes the source from the agent; no grounding check in `AgentFlight::take_reply`; each origin permitted; an empty fault line.

## Decisions for the owner

1. **Evidence baseline.** `coordinator/tests.rs` is bound by digest in `docs/control/evidence-graph.evg`, so the two input-source results are recorded at branch commit `0b124860`. A squash merge makes that commit unreachable. Record the two results again at the durable commit before the merge, as the merge and rebase rule of the graph states.
2. **Size caps.** Five browser modules were at their caps. The caps move by the lines that this change adds (`control-loop.js` 952→962, `control-shell.js` 341→372, `cockpit-readout.js` 763→764, `main.js` 239→279, `viewer-elements.js` 16→23).
3. **Focus rule.** The agent in the browser flies only while the window has the focus, as a person's input does. A headless agent that must fly with no window is the headless port.
4. **Apple client.** It links the same control runtime, so the agent input source is there. It has no agent panel and no gateway connection.
5. **New dependency.** `httparse` (no dependencies of its own) in `tools/intent-pilot`.


luofang34:
## Two independent reviews of this PR, and what changed

One reviewer read the client module (control runtime, browser modules, evidence graph); another read the headless port, the adapters, the harness and the gateway. They found 2 blockers, 12 should-fix items and a set of nits. Each is answered below; the fixes are the commits from `5fbb935` on this branch, and each fix has a test that failed before it.

### Client module

| # | Finding | Answer |
| --- | --- | --- |
| 1 | **Blocker.** A flight-scope toggle under an engaged agent made the shell read the agent's velocity demand as an attitude setpoint; the feel selector and the reset button also bypassed the agent. | Fixed. The control loop releases the agent before a flight-scope or mode change, a simulation reset, a feel request and the release of control authority. A test drives each path. |
| 2 | The agent survived every authority loss, and "Resume control" resumed it. | Fixed by the same release, and a new session starts with the operator as the source; the browser module follows the runtime when it no longer has the agent. |
| 3 | The safety-press case was vacuous (the handover tick never arms). | Fixed. The test holds the press through the install and the first live ticks, and proves that a release and a new press is the operator's own arm. |
| 4 | The keyboard half of the override was untested. | Fixed. A keyboard test and a gimbal-reset-press test cover it; both are in the evidence graph. |
| 5 | `engage()` read its conditions before the waits. | Fixed. The conditions are read again after each wait; a second engage while one is under way is refused; a refused engage frees the runtime object. |
| 6 | The gimbal reset press did not release the agent. | Fixed. |
| 7 | The browser gave up at 30 s while the gateway waits 120 s for its adapter. | Fixed. The browser waits 125 s. |

### Headless port, adapters, harness, gateway

| # | Finding | Answer |
| --- | --- | --- |
| 1 | **Blocker.** `quit`, an end of input or a verdict during the disarm sent a return to base to a vehicle on the ground, and it took off. | Fixed in the executor (a return on the ground or during the disarm is refused) and in the pilot (the refusal ends the run; the recovery starts only from a phase in the air). |
| 2 | A late model reply or a leftover scripted message replaced the return to base after `quit`. | Fixed. After the operator ends the run, replies are recorded and not flown, and the script releases nothing. |
| 3 | No request/reply correlation on the line protocol; a native print into fd 1 became a reply. | Fixed. Each request carries a number, the adapters echo it, a reply to an earlier request is skipped, and a reply to a number never sent is a fault. The adapters move fd 1 to stderr. Two tests drive the port with a shell adapter. |
| 4 | Gateway shutdown was unbounded and Ctrl-C was disarmed. | Fixed. A grace time, then the open requests and the adapter task are aborted. |
| 5 | `--listen 0.0.0.0` has no authentication; 32 MiB was allocated per connection before bytes arrived. | Partly. The origin is checked before the body is read, the body grows as it arrives, at most 16 connections are served, and a listen beyond loopback logs a warning. Authentication is not added; the README names the limit. |
| 6 | The README's "first run" wording did not match the ledger (the held-out suite ran twice, and the grounding rules were written after seeing held-out output). | Fixed. The README states the runs plainly and reports a third run with the checks as they are. The correct counts equal the first run for each adapter. |
| 7 | An unparseable directive kind became `unable`, which scores correct on chatter cases. | Fixed. It is a fault now. |
| 8 | The close after a yield or a release could drop the release with the connection. | Fixed. The close waits up to 2 s for the host's next word on authority. |
| 9 | Committed Python bytecode. | Removed; `__pycache__/` is ignored. |
| nits | untyped gateway CLI error; `--help` on stderr; `asks.send` on a full queue; the stdin task at exit; the ledger key; near-template held-out cases; the takeoff messages of two scenarios ride the implicit "flew" rule. | Left as follow-ups. |

The reviewers confirmed: engage and release go through the swap with an advanced revision and a retained lease; the announcement follows the active source; a pad stick or a safety press releases the agent with a neutral handover; the demand is consumed in one tick and bounded; the single publish gate and the single lease encoder are untouched; the evidence blobs and artifact digests match the tree; no pin is unpinned; the scope is chosen by name; frames stop when the lane is gone.

The evidence results were recorded again at the fix commit, and the hard gate passes with the pending review slot only.

PR #600, opened 2026-09-17
feat(planning): add the mission planner and navigation search (workstation work, Swift unverified)

luofang34:
## What this is

This draft publishes the mission planner and navigation search as they stand on the workstation. The work was uncommitted there; this PR groups it into five commits and does not change it.

- `pilotage-planning`: navigation index and search, routes (distance, track, elapsed and arrival time), vehicle profiles, and coordinated missions (assignments, responsible persons, swarms, times on target) with validation.
- The Swift binding for that crate. Swift does not calculate a route or validate a mission.
- An attitude cage in `PilotageCore` (level-offset correction) with tests.
- The planner and search in the Apple app, with unit tests and a UI test target.
- Documents: implemented scope, route editor workflows, an example vehicle profile library, and the X-Plane planner check of 2026-09-13.

## Verification

- `cargo check --workspace`, 47 tests on the touched crates, clippy with `-D warnings`, and `cargo fmt --check` pass on this head.
- **Not verified:** all Swift code and the UI tests.
- `Cargo.lock` is not updated for `pilotage-planning` in this branch. The workstation lockfile blended entries from more than one lane, so it was left out. Regenerate it after the rebase.

## Left out on purpose

The workstation also held 76 uncommitted flight-tune files. A comparison with `main` showed 48 equal an older `main` version, 12 equal the current one, and 16 match no version. They are kept on a local branch (`parked/flight-tune-worktree-leftovers`) and are not proposed.

Stacked on #599 and #598. It inherits their blockers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #609, opened 2026-09-18
docs(adr): keep the GDL 90 codec in a crate of its own

luofang34:
## Summary

The GDL 90 codec now has a crate of its own: `gdl90` in `luofang34/Gdl90` (private for now; see the decision below). This record states why, what the crate holds, the dependency direction, and the two follow-up adapter crates.

The problem it answers: the Apple client of #600 imports GDL 90 types from `aero_link::core`, and the pinned AeroLink revision does not have them. They exist only in the AeroLink working tree. With the codec in its own crate, the Apple client reads GDL 90 with the pinned AeroLink revision, and the AeroLink pin moves for receiver changes only.

## The crate

`gdl90` holds the wire format and nothing else: the frame with its CRC and byte transparency, and one encoder and one decoder for each message that an appliance sends (Heartbeat, Uplink Data, Ownship Report, Ownship Geometric Altitude, Traffic Report, Stratux AHRS, ForeFlight extension AHRS). It is `no_std` with no I/O, no clock, no allocation and no domain model. 36 tests, with the published examples of the specification byte for byte; CI runs fmt, clippy, tests, strict docs and no_std builds for a Cortex-M0, a RISC-V and a wasm target.

The encoders and the traffic decoder come from the AeroLink working tree. The other decoders and the traffic encoder are new, so that both sides of the interface are in one place.

## Decisions for the owner

1. **Visibility.** The repository is private. The Pilotage CI job that builds the Apple binding must fetch it: either make `Gdl90` public (it is a codec of a public specification, with nothing sensitive), or add a deploy key for it to the Pilotage CI secrets as for Airmass and Surveillance.
2. **AeroLink.** The AeroLink working tree still holds `gdl90*.rs` in `aero-link-core`. The record proposes that AeroLink takes the crate and drops its copies. I did not touch that tree.
3. **Follow-ups.** `surveillance-gdl90` and `airmass-gdl90` in their repositories, so that the domain mappings leave the Pilotage binding.


luofang34:
Correction to the description: the `Gdl90` workflow is defined (fmt, clippy, tests, strict docs, no_std builds for Cortex-M0, RISC-V and wasm), but GitHub has not run it. Each attempt ends with "The job was not started because recent account payments have failed or your spending limit needs to be increased" — the repository is private, so its minutes are billed, while Pilotage and Indicate are public. The same gates pass on claw (`~/Gdl90`, `cargo test` 36 tests + doctest, `scripts/check-no-std.sh`). Making the repository public removes both this and the deploy-key question in the decision list.

